PR #4, opened 2019-07-05
Strip down Reveal Frontend to a basic skeleton.

### LINKED ISSUE #1: Fork/copy reveal-frontend
Currently we have an empty repo an we need to populate it with the onaio/reveal-frontent codebase.

### LINKED ISSUE #2: Identify and Remove Reveal Specific Code
If copying code straight from reveal-frontend then we need to remove code specific to the context of Reveal (components, services, vars, etc).

p-netm:
closes #1 , closes #2 

Remove Reveal specific code from reveal-frontend


src/configs/__mocks__/env.ts
```diff
@@ -0,0 +1,11 @@
+/** You can hard code variables that would have been
+ * extracted from the environment here, and then mock
+ * this file from your test file.
+ * The goal is to have consistent test results
+ */
+
+export const SUPERSET_PLANS_SLICE = 0;
+export type SUPERSET_PLANS_SLICE = typeof SUPERSET_PLANS_SLICE;
+
+export const GISIDA_ONADATA_API_TOKEN = 'hunter2';
```
moshthepitt: I think that this one should be removed

src/setupTests.ts
```diff
@@ -0,0 +1,23 @@
...
+import { faUser } from '@fortawesome/free-regular-svg-icons';
+import { faMap } from '@fortawesome/free-solid-svg-icons';
+import enzyme from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+
+enzyme.configure({ adapter: new Adapter() });
+library.add(faMap);
+library.add(faUser);
+
+window.maps = [
```
moshthepitt: I think this is meant for Gisida and should probably be removed.

src/constants.tsx
```diff
@@ -0,0 +1,39 @@
+// strings
```
moshthepitt: I think many of the contants in line 1 - 23 are not used elsewhere.  Can you remove the ones that are not used, please?

PR #9, opened 2019-07-09
Reducer and ui component for opensrp clients endpoint

### LINKED ISSUE #6: Create sample OpenSRP service
Similar to #5, we need to create a service that pulls data from the identified endpoint and then populates the redux store.

- [x] Pull a list of clients
- [x] Populate redux store
- [x] Display a list of clients on a clients page
- [x] POST a client to the OpenSRP server

p-netm:
closes #9 , closes #6


src/components/page/Header/Header.tsx
```diff
@@ -86,6 +86,11 @@ export class HeaderComponent extends React.Component<HeaderProps, State> {
                   </NavLink>
                 </NavItem>
               )}
+              <NavItem>
+                <NavLink to={CLIENT_URL} className="nav-link" activeClassName="active">
+                  clients
```
cKellyDesign: @p-netm let's make this a capital `C` to match the casing of the other links

src/helpers/utils.tsx
```diff
@@ -23,3 +24,18 @@ export function oAuthUserInfoGetter(apiResponse: { [key: string]: any }): Sessio
     }
   }
 }
+
+/** func to only extract bits the client values needed from API response */
+export function extractClient(rawClient: FlexObject): Client {
```
cKellyDesign: @p-netm I like this functionality a lot! Do you think it makes sense to move this type of extractor into the Clients dux module? Maybe not since this app is what's responsible for handing Client objects to the reducer module.

cc @moshthepitt - thoughts?

src/containers/Clients/List/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  public render() {
+    /** render loader if there are no clients in state */
+    const { clientsArray } = this.props;
+    if (!some(clientsArray)) {
+      return <Loading />;
+    }
+    return (
+      <div>
+        <table className="table table-shadow">
```
cKellyDesign: @p-netm Since we're using Reactstrap, let try and use their [components](https://reactstrap.github.io/components/tables/) when possible to get consistent / built-in bootstrap styling without needing to worry about `className`s

src/containers/Clients/List/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+          <tbody>
+            <tr>
```
cKellyDesign: I think this `<tr />` should be in `<thead />`

src/helpers/utils.tsx
moshthepitt: thats a good question @cKellyDesign.  In reveal we have an extractPlan used slightly differently but which lives in the utils.  I can certainly see why we may want to move it into the ducks module for both cases

src/components/page/Header/Header.tsx
p-netm: ok, will do; done

src/helpers/utils.tsx
p-netm: ```extractclient``` moved to ```src/store/ducks/clients.ts```

src/services/clients/index.ts
```diff
@@ -0,0 +1,39 @@
+import { OPENSRP_BASE_API_ENDPOINT, OPENSRP_CLIENT_ENDPOINT } from '../../configs/env';
+import { FAILED_TO_RETRIEVE_CLIENTS, SERVICE_HEADERS } from '../../constants';
+
+class ClientService {
+  private CLIENTS_API_ENDPOINT = `${OPENSRP_BASE_API_ENDPOINT}/${OPENSRP_CLIENT_ENDPOINT}/search`;
+
+  public async getClientsList() {
+    interface Response {
+      data: any;
```
moshthepitt: I am hoping that if we write a type definition for the raw OpenSRP client response then we can avoid using `any` here.

src/store/ducks/clients.ts
```diff
@@ -0,0 +1,126 @@
...
+export const reducerName = 'clients';
+
+/** Interface for client object */
+export interface Client {
+  id: string;
+  firstName: string;
+  gender: string;
+  lastName: string;
+  lastContactDate: string;
+  location: string;
```
moshthepitt: What does this field represent?

src/store/ducks/clients.ts
```diff
@@ -0,0 +1,126 @@
...
+ * @return {Client | null} a client obj if the id is found else null
+ */
+export function getClientById(state: Partial<Store>, id: string): Client | null {
+  return get(getClients(state), id) || null;
+}
+
+// utils for clients dux module
+
+/** func to only extract bits the client values needed from API response */
+export function extractClient(rawClient: FlexObject): Client {
```
moshthepitt: I am hoping that if we write a type definition for the raw OpenSRP client response then we can avoid using `FlexObject` here.

tslint.json
```diff
@@ -1,6 +1,9 @@
 {
   "extends": ["tslint:recommended", "tslint-react", "tslint-config-prettier"],
   "rulesDirectory": ["tslint-plugin-prettier"],
+  "linterOptions": {
+    "exclude": ["src/services/clients/tests/fixtures.ts"]
```
moshthepitt: Can we add the linter exclusions in the module in question instead of here please?

src/services/clients/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+    let result: Response | null = null;
+
+    await fetch(this.CLIENTS_API_ENDPOINT, { headers: { ...SERVICE_HEADERS }, method: 'GET' })
+      .then(async response => {
+        const returnObj = await response.json();
+        let data: Array<{ [key: string]: any }> = returnObj;
+        // check that the returned data is an array, if there was an
+        // error this will usually comeback as an object
+        if (!Array.isArray(returnObj)) {
```
moshthepitt: Will we return an error message in case we end up here?

p-netm: yes so the idea is to have a standard interface for the service Response irregardless of the api response.

src/configs/__mocks__/env.ts
```diff
@@ -4,5 +4,10 @@
  * The goal is to have consistent test results
```
moshthepitt: Can we mock the `REACT_APP_WEBSITE_NAME` env var please?

p-netm:
Messed up my history pretty badly, had to to redo this Pr

PR #15, opened 2019-07-23
Clients reducer, service and clientsList ui component

### LINKED ISSUE #5: Create sample reducer module
We need to identify one of the OpenSRP endpoints and implement a sample reducer module for it.

### LINKED ISSUE #6: Create sample OpenSRP service

p-netm:
Adds: A sample Clients reducer, a service to feed the reducer with data from an api, and a ui component to display the data.

closes #6 , closes #9 , closes #5


src/services/clients/index.ts
```diff
@@ -0,0 +1,13 @@
+import { OPENSRP_CLIENT_ENDPOINT } from '../../configs/env';
+import { OpenSRPService } from '../../services/opensrp';
+
+class ClientService {
```
moshthepitt: Is this really necessary?

It seems we could replace this with something like:

```
const clientService = new OpenSRPService('clients/search');
// then to get the list do
clientService.list()
```

p-netm: Removed the clientsService totally, components now make api calls directly via the opensrpService inside of the component.

Issue #25, opened 2019-08-22
Create a pagination component


redoy11:
Issue #21 includes a pagination bar at the bottom of the page. This issue involves creating a pagination bar component that can be used in Issue #21 as well as in other views. 
To resolve the issue:

- [ ] create a prop that can be used to highlight the active page.

- [ ] create a prop that can be used to show the list of pages in the bar

- [ ] create a handler prop that will be triggered on selecting pages

Issue #22, opened 2019-08-16
Household Profile View


hmakandi:
The design for the household profile view is as below
<img width="1026" alt="Screenshot 2019-08-16 at 13 23 22" src="https://user-images.githubusercontent.com/18394882/63161705-5e4ff000-c029-11e9-9fa8-85203a152127.png">

- [ ] Have the link that goes back to the Household register view ([Issue 21](https://github.com/OpenSRP/opensrp-web/issues/21))

- [ ] In the basic information card, on click of the edit Button, make the fields Family Name, Head of household, and phone number

- [ ] The change of Head of Household in the above edit should show a drop down of all family members, such that the head of the household is a family member

- [ ] The edit should make changes to both OpenSRP and OpenMRS

- [ ] In the table card that shows the family members, actions available is view which shows the members profile page, and remove, which removes that member from that family, but not from the register. The member should remain and exist in the all clients register.

- [ ] Pagination should be enabled should the members exceed the configured page count(currently at 50 and should be configurable) 
<img width="953" alt="Screenshot 2019-08-16 at 13 33 57" src="https://user-images.githubusercontent.com/18394882/63162902-6eb59a00-c02c-11e9-9430-712aa2c4494c.png">

PR #26, opened 2019-08-22
Create Pagination Component

redoy11:
Resolves Issue #25.

Created a pagination component that can be used in Issue #22. Props are as mentioned in the Issue #25.


p-netm:
will use a library that provides the pagination component.

Issue #24, opened 2019-08-22
Clients Dux reducer gets written wholly each time new clients are added to it


p-netm:
__The Current Behaviour__:
The newly fetched client records overwrite the existing client records inside the store.

__The expected Behaviour [_For the existing clients dux module and all other future modules_ ]__:

That once new clients are added to the store via a dispatch action, the new clients are added to the already existing clients records

PR #28, opened 2019-08-23
WIP - Miecd Vietnam

HusnaHariz:
Web app for MIECD-Vietnam project


src/store/ducks/sms_events.ts
```diff
@@ -0,0 +1,88 @@
...
+  };
+  //   console.log("gets called", actionCreated);
+  return actionCreated;
+};
+
+// The reducer
+
+/** interface for sms state in redux store */
+// tslint:disable-next-line: class-name
+interface smsStateReducer {
```
p-netm: Use PascalCase for type names

src/store/ducks/sms_events.ts
```diff
@@ -0,0 +1,88 @@
...
+/** the sms reducer function */
+export default function reducer(
+  state: smsStateReducer = initialState,
+  action: TestActionTypes
+): smsStateReducer {
+  switch (action.type) {
+    case FETCHED_SMS:
+      return {
+        ...state,
+        payload: action.payload,
```
p-netm: see this issue openSRP/opensrp-web#24 for context on the problem here, the issue is fixed in   openSRP/opensrp-web#31. The specific fix of interest is in this [commit](https://github.com/OpenSRP/opensrp-web/pull/31/commits/929e05e00f072701cb68d88872149ea9809819de)

src/components/RiskColoring/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+const riskColoring = (props: RiskColoringProps) => {
+  const { Risk } = props;
+  switch (Risk) {
+    case 'No Risk':
+      return (
+        <td className="green" id="default-width">
+          {Risk}
+        </td>
+      );
+    case 'no risk':
```
p-netm: the code for the first 2 cases is quite indetical, maybe a better approach would be to standardize the the Risk prop, say change it to lowerCase or uppercase.

or have a look up table to decide only those parts that are unique and then return a single jsx instance; something like:

```typescript
  /** Interface for the risk look up dictionary Object */
  interface ClassNameAndTextLookup {
    [key: string]: { className: string; label: string };
  }

  // standardize the Risk variable's string by changing case
  const upperCasedrisk = Risk.toUpperCase();

  /** lookup dictionary for different Risk categories */
  const classNameAndTextLookup: ClassNameAndTextLookup = {
    HIGH: { className: 'yellow', label: 'High risk' },
    LOW: { className: 'green', label: 'Low Risk' },
    'NO RISK': { className: 'green', label: 'No Risk' },
    RED: { className: 'red', label: 'Red Alert' },
    DEFAULT: { className: 'risk-default', label: Risk },
  };

  // use default if given Risk definition was not found in our dictionary
  const risk = classNameAndTextLookup.upperCasedRisk ? upperCasedrisk : 'DEFAULT';

  return (
    <td className={classNameAndTextLookup[risk].className} id="default-width">
      {classNameAndTextLookup[risk].label}
    </td>
  );
}
```

src/containers/LogFace/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    supersetFetch('1277').then((result: any) => {
+      fetchTestDataActionCreator(result);
+    });
+  }
+
+  // tslint:disable-next-line: no-empty
+  public handleSubmit() {}
+  // tslint:disable-next-line: no-empty
+  public handleTermChange() {}
+  public handleKeyPress(event: { key: string }) {
```
p-netm: React has types for DomEvents.

src/configs/settings.ts
```diff
@@ -43,4 +43,13 @@ if (ENABLE_ONADATA_OAUTH) {
   };
 }
 
+export const Types = [
```
moshthepitt: Can you give this a better name please?  What 'Types' are these?  Keep in mind that OpenSRP web will be used in very many ways so all variables need to be as descriptive as possible.

src/containers/LogFace/index.tsx
```diff
@@ -69,41 +94,46 @@ export class LogFace extends React.Component<PropsInterface, {}> {
...
+                        {location}
+                      </DropdownItem>
+                    );
+                  })}
+                </DropdownMenu>
+              </Dropdown>
             </div>
             <div className="location-type-filter">
               Type
-              <Form onSubmit={this.handleSubmit}>
```
moshthepitt: Please use the Formik library like we do everywhere else.

src/containers/LogFace/index.tsx
```diff
@@ -2,59 +2,84 @@
...
-      this.handleSearch();
+  public componentWillReceiveProps(nextProps: any) {
+    if (!this.props.testData.length) {
+      this.setState({
+        filteredData: nextProps.testData,
+      });
     }
   }
+
+  public handleSubmit(e: any) {
```
moshthepitt: If you use something like Formik you dont have to do this.

src/containers/pages/Home/Home.tsx
```diff
@@ -31,7 +33,7 @@ class Home extends React.Component<{}, {}> {
                   <p>Display of all messages receioved from MIECD in chronological order</p>
                 </div>
                 <div id="cont-sized">
-                  <a href="/log-face">
+                  <a href={LOGFACE_URL}>
```
moshthepitt: Is it possible to make a prop on the Home component for this?

src/components/RiskColoring/index.tsx
FrankApiyo: I went with making all the risk prop lowercase

Issue #34, opened 2019-08-27
Extend Clients Redux for Households


redoy11:
- [ ] Add action creators for Households fetching
- [ ] Add action creators for removal of Households from store
- [ ] Modify clients dux to store households by Id to separate between other types of clients and families
- [ ] Add household specific selectors 

@redoy11

_Originally posted by @p-netm in https://github.com/OpenSRP/opensrp-web/issues/27#issuecomment-525170285_

p-netm:
closed by #37

PR #35, opened 2019-08-28
Issue 34

redoy11:
Reference to Issue #34, the patch extends client dux to accommodate Households in store. It also adds household-specific action and selectors

### DIFF
--- a/src/store/ducks/clients.ts
+++ b/src/store/ducks/clients.ts
@@ -74,6 +74,7 @@ export type ClientsActionTypes =
   | FetchClientsAction
   | FetchHouseholdsAction
   | RemoveClientsAction
+  | RemoveHouseholdsAction
   | AnyAction;
 
 // action Creators
--- a/src/store/ducks/clients.ts
+++ b/src/store/ducks/clients.ts
@@ -172,3 +172,11 @@ export function getClientsArray(state: Partial<Store>): Client[] {
 export function getClientById(state: Partial<Store>, id: string): Client | null {
   return get(getClients(state), id) || null;
 }
+
+/** returns all households in the store as values whose keys are their respective ids
+ * @param {Partial<Store>} state - the redux store
+ * @return { { [key: string] : Household} } - households object as values, reepective ids as keys
+ */
+export function getHouseholdsById(state: Partial<Store>): { [key: string]: Household } {
+  return (state as any)[reducerName].householdsById;
+}
--- a/src/store/ducks/clients.ts
+++ b/src/store/ducks/clients.ts
@@ -101,6 +101,7 @@ export const removeClientsAction = {
 /** interface for clients state in redux store */
 interface ClientState {
   clientsById: { [key: string]: Client };
+  householdsById: { [key: string]: Household };
 }
 
 /** Create an immutable clients state */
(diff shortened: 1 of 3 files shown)


src/store/ducks/clients.ts
```diff
@@ -1,4 +1,5 @@
 import { get, keyBy, keys, values } from 'lodash';
+import { type } from 'os';
```
p-netm: Remove unused imports

src/store/ducks/tests/clients.test.ts
```diff
@@ -65,4 +72,27 @@ describe('reducers/clients', () => {
...
+    });
+    expect(getHouseholdsArray(store.getState())).toEqual(
+      values([fixtures.household1, fixtures.household2])
+    );
+    expect(getHouseholdById(store.getState(), '1bcb682a-0f31-4935-9114-c4d33d148617')).toEqual(
+      fixtures.household1
+    );
+  });
+
+  it('Adds new households to store instead of overwriting existing ones', () => {
```
p-netm: We probably dont want information from a test case seeping into other testCases. Lets create an action to remove households from store before each testCase. Leads to more predictable and easier to think about changes to the test file.

p-netm:
Great work.

Made a few changes that i think i am responsible for, so hope you don't mind.

src/store/ducks/tests/clients.test.ts
redoy11: Okay. Creating an action to remove households and changing the test cases now.  Also updating Issue #34 to include this.

p-netm: Great. now lets use the remove households to ensure that each test case starts on a clean page. This can be easily done by adding the now tested `removeHouseholdsAction`  in the `beforeEach` block

redoy11: Should I add the removeHouseholdsAction at the start of all the test blocks related to households or just adding on the last two test blocks will suffice? I think adding on the last two test blocks will be better.

p-netm: let add it [here](https://github.com/OpenSRP/opensrp-web/blob/f3773efdd1191c9df639c5dc41ccf013f165ba7f/src/store/ducks/tests/clients.test.ts#L26) both remove actions for clients and households

redoy11: Done

src/store/ducks/clients.ts
```diff
@@ -64,6 +88,15 @@ export const fetchClients = (clientsList: Client[] = []): FetchClientsAction =>
   type: CLIENTS_FETCHED,
 });
 
+/** Fetch households action creator
```
moshthepitt: Does `fetchHouseholds` really take two params?

redoy11: fetchHouseholds take one parameter and that is the household list.

moshthepitt: I think there might be an error with this docstring which seems to say two params?

redoy11: Got the error.  The last param mentioned in the docstring will actually be the return type. Fixing it now.

redoy11:
Merging is not possible since this PR contains commits without signing. So closing this thread.

PR #43, opened 2019-09-03
Header Modification and Addition of Side Menu Bar

### LINKED ISSUE #40: Modifying the header and styling to look as in mockups
The header should look like this:
Possible file changes to(NB: not an exhaustive list) :
- the app component
- header component
- and both their styling

This is the goal:
image

- Responsiveness is key

redoy11:
This fixes #40.
The patch contains the following changes:

- Adds side menu component.

- Shifts the extra nav items from the header to the side menu bar.

- Modifies the style of header according to the mockup.


src/components/page/Header/Header.css
```diff
@@ -15,3 +12,27 @@
...
+
+.bg-navbar{
+  background: #2B475E !important;
+  box-shadow: 0 1px 12px 0 rgba(195, 203, 222, 0.59);
+  padding: 0;
+}
+
+.bg-top-nav{
+  background: #233F56;
+  box-shadow: 0 1px 12px 0 rgba(195, 203, 222, 0.59);
```
p-netm: This style is not necessary. the box shadow is not part of the mockup

p-netm:
This section in the header is not quite similar to the mockup.

__Expected__
expected

__current look__
got

notice the darker blue color boundary. also, this may not have in the mockup but lets change the color of the account owner text to something other than the default blue during hover, it should be a lighter grey colour

src/components/page/SideMenu/SideMenu.tsx
```diff
@@ -0,0 +1,94 @@
...
+  parentNav: NavCollapseObj;
+  childNavs: NavObj[];
+}
+
+export interface SideMenuState {
+  isActive: string[];
+  isExpand: string[];
+}
+
+const sideMenuState = {
```
p-netm: yeah, we have an interface we just didn't use it on line here

src/components/page/SideMenu/SideMenu.tsx
```diff
@@ -0,0 +1,94 @@
...
+import SubMenu, { NavCollapseObj, NavObj, SubMenuProps } from '../SubMenu/SubMenu';
+import './SideMenu.css';
+
+export interface NavModuleObj {
+  identifier: string;
+  parentNav: NavCollapseObj;
+  childNavs: NavObj[];
+}
+
+export interface SideMenuState {
```
p-netm: `isExpand` is kinda confusing, given that is-prefixed variables are expected to contain a boolean value, but here the value is an array of string.

src/components/page/SubMenu/tests/SubMenu.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+      parentNav: { navIcon: ['fas', 'cog'], navLabel: 'Admin' },
+      setSideMenuActive: null,
+      setSideMenuToggle: null,
+    };
+    const wrapper = mount(
+      <Router history={history}>
+        <SubMenu {...props} />
+      </Router>
+    );
+    expect(toJson(wrapper.find('SubMenu'))).toMatchSnapshot();
```
p-netm: Maybe we can reduce the the snapshots further with more targeted snapshots. like say you could check that we have a navItem for `Admin`

src/components/page/SideMenu/SideMenu.tsx
p-netm: `isActive` too

src/components/page/SubMenu/tests/SubMenu.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { createBrowserHistory } from 'history';
+import React from 'react';
+import { Router } from 'react-router';
+import SubMenu, { SubMenuProps } from '../SubMenu';
+library.add(faUser);
+library.add(faCog);
+
+const history = createBrowserHistory();
+
+describe('components/page/SideMenu', () => {
```
p-netm: Probably meant ``components/page/SubMenu``

src/components/page/SideMenu/tests/SideMenu.test.tsx
```diff
@@ -0,0 +1,38 @@
+import { library } from '@fortawesome/fontawesome-svg-core';
+import { faUser } from '@fortawesome/free-regular-svg-icons';
```
p-netm: It would be awesome if we put the fontawesome stuff for the tests inside the setupTest.ts file. It will help keep the code dry.

src/components/page/SideMenu/tests/SideMenu.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    );
+  });
+
+  it('renders header correctly', () => {
+    const wrapper = mount(
+      <Router history={history}>
+        <SideMenu />
+      </Router>
+    );
+    expect(toJson(wrapper.find('SideMenu'))).toMatchSnapshot();
```
p-netm: this snapshotMatcher is most likely to lead to a huge snapshot as evidenced in the resulting snap file.

src/components/page/SideMenu/tests/SideMenu.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  it('renders without crashing', () => {
+    shallow(
+      <Router history={history}>
+        <SideMenu />
+      </Router>
+    );
+  });
+
+  it('renders header correctly', () => {
```
p-netm: please pay attention to places where you copy paste code snippets, make sure to recheck everything including docstrings and and string literals, and refactor where necessary

src/components/page/SideMenu/tests/SideMenu.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const wrapper = mount(
+      <Router history={history}>
+        <SideMenu />
+      </Router>
+    );
+    expect(toJson(wrapper.find('SideMenu'))).toMatchSnapshot();
+    wrapper.unmount();
+  });
+});
```
p-netm: I feel like this component could use some tests, for instance we have a few conditionally rendered subcomponents, we also have clickhandlers, and we should be certain that all this work as expected

.circleci/config.yml
```diff
@@ -14,8 +14,17 @@ jobs:
           REACT_APP_WEBSITE_NAME: "OpenSRP"
           REACT_APP_DOMAIN_NAME: "http://localhost:3000"
           REACT_APP_THEME_COLOR: "000000"
-          REACT_APP_ENABLE_USERS: "true"
-          REACT_APP_ENABLE_ABOUT: "true"
+          REACT_APP_ENABLE_CLIENT_RECORDS_MODULE: "true"
```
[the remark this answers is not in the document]
redoy11: It does make more sense. I will change it like that.

src/components/page/SubMenu/SubMenu.tsx
```diff
@@ -0,0 +1,92 @@
...
+              </Nav>
+            );
+          })}
+        </Collapse>
+      </div>
+    );
+  }
+
+  private toggle = () => {
+    this.props.setSideMenuToggle(this.props.identifier);
```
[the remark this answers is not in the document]
p-netm: > `this` would be undefined inside this 2 methods the `toggle` and `setActive...` [see here](https://reactjs.org/docs/handling-events.html).
> 
> > You have to be careful about the meaning of this in JSX callbacks. In JavaScript, class methods are not bound by default. If you forget to bind this.handleClick and pass it to onClick, this will be undefined when the function is actually called.
> 
> I would also like to point out that this shows the testing might be wanting, specifically for this case, you need to simulate a user performing some action that would lead to this methods getting called and then checking that the methods were actually called. This is of course in addition to checking that the functions actually do what they are supposed to i.e. the `set...` functions were called with the expected/correct arguments.

My bad; I wasn't paying enough attention to realize you were using `public class fields syntax`, and thus `this` would be automatically bound to the enclosing scope. Nice approach on that, however we still need to add more tests to cover and not limited to, the mentioned cases.

.circleci/config.yml
```diff
@@ -14,8 +14,17 @@ jobs:
...
+          REACT_APP_ENABLE_CLIENT_PAGE: "true"
+          REACT_APP_ENABLE_HOUSEHOLD_PAGE: "true"
+          REACT_APP_ENABLE_ANC_PAGE: "true"
+          REACT_APP_ENABLE_CHILD_PAGE: "true"
+          REACT_APP_ENABLE_REPORT_MODULE: "true"
```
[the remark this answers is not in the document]
redoy11: Currently, the Report module does not have any child pages. How should I proceed with this?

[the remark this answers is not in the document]
p-netm: ooh sorry i meant to flag this for the `ADMIN pages`, i think the report module can come from the env variables.

src/components/page/SubMenu/SubMenu.tsx
```diff
@@ -0,0 +1,125 @@
...
+/** interface for collapsable Nav Menu Labels
+ * @property {string} navLabel - text label visible on menu item
+ * @property {IconProp} navIcon- fontawesome icon present on menu item
+ */
+export interface NavCollapseObj {
+  navLabel: string;
+  navIcon: IconProp;
+}
+
+type SetSideMenuToggle = (identifier: string) => void;
```
p-netm: why do we need this function as a prop, is there any variations as to how we toggle the submenu, what if this was solely handled by the component as an uncontrolled action.

src/components/page/SubMenu/SubMenu.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+export interface SubMenuProps {
+  identifier: string;
+  isCollapseMenuActive: boolean;
```
p-netm: lets name this `colapseMenu` . but before that i would like to know the differnce in how `isCollapseMenuActive` and `isExpand` are used. seems like they refer to the same thing.

src/components/page/SubMenu/SubMenu.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  private toggle = () => {
+    const { identifier, setSideMenuToggle } = this.props;
+    if (setSideMenuToggle) {
+      setSideMenuToggle(identifier);
+    }
+  };
+
+  private updateCollapseMenuStatus = () => {
```
p-netm: Please make sure your functions have docstrings

src/components/page/SubMenu/SubMenu.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const { childNavs, location } = this.props;
+    let isCurrentURLinChildPresents: boolean = false;
```
p-netm: this variable name has 6 words in it. can we make it shorter while maintaining its expressiveness

src/components/page/SubMenu/SubMenu.tsx
```diff
@@ -0,0 +1,125 @@
...
+  isCollapseMenuActive: false,
+  isExpand: true,
+};
+
+class SubMenu extends React.Component<SubMenuProps & RouteComponentProps, SubMenuState> {
+  public static defaultProps = defaultSubMenuProps;
+  constructor(props: SubMenuProps & RouteComponentProps) {
+    super(props);
+    this.state = {
+      isCollapseMenuActive: this.props.isCollapseMenuActive,
```
p-netm: `isCollapseMenuActive` do you think its a good idea to have this as controlled and uncontrolled data.
if the value from the prop serves only the purpose of providing a default value, why dont we scrap that and cactually define it in the default props only.

src/components/page/SubMenu/SubMenu.tsx
```diff
@@ -0,0 +1,125 @@
...
+    };
+  }
+  public componentDidMount() {
+    this.updateCollapseMenuStatus();
+  }
+
+  public componentDidUpdate() {
```
p-netm: quick questino here, why do we have to call the same function in both this lifecycle hooks

src/components/page/SubMenu/tests/SubMenu.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+      isExpand: true,
+      parentNav: { navIcon: ['fas', 'cog'], navLabel: 'Admin' },
+    };
+    const wrapper = mount(
+      <Router history={history}>
+        <SubMenu {...props} />
+      </Router>
+    );
+    /** match Users navItem json */
+    expect(toJson(wrapper.find('NavItem .nav-item-extend').first())).toMatchSnapshot();
```
p-netm: is there a way to make this test more targeted and inclined towards the statement that describes it, i.e for this case we want the test to show that a link with the text `users` and another with `roles` were shown on the page and not just any other link

src/components/page/SubMenu/tests/SubMenu.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <SubMenu {...props} />
+      </Router>
+    );
+    wrapper.find('Nav .side-collapse-nav').simulate('click');
+    expect(mockCallBack.mock.calls.length).toEqual(1);
+    wrapper.find('Nav .side-collapse-nav').simulate('click');
+    expect(mockCallBack.mock.calls.length).toEqual(2);
+    wrapper.unmount();
+  });
+});
```
p-netm: lets add a test that checks whether the css classes are dynamically as populated. Am thinking 
simulate a click on one of the chiild navigation links. check that the active class has been set for 
the parent link and the child nav. click another child  nav, then check that the previously clicked child nave does not have the active class anymore and the newly clicked nav has the active class.

This will also raise the question of how you are handling child nav clicks across different parents and how they affect each other active states. like say the case shown in the below screenshot should have been flagged by a test like this.

src/components/page/SubMenu/SubMenu.tsx
p-netm: `isCurrentURLinChildPresents` is there a typo in this, cause i dont know exactly what value is being referred to here. o rmaybe thee intention was `isCurrentURLinChildPresent`?

src/components/page/SubMenu/tests/SubMenu.test.tsx
p-netm: bug

p-netm:
ill be posing some architectural considerations later .

src/App/App.tsx
```diff
@@ -11,22 +12,28 @@ import { providers } from '../configs/settings';
...
     return (
-      <Container>
+      <Container fluid={true}>
         <ConnectedHeader />
         <Row id="main-page-row">
-          <Col>
+          <Col md={2} className="side-menu-section">
+            <SideMenu />
+          </Col>
+          <Col md={10} className="contaioner-section">
```
p-netm: typo in `contaioner-section`, make test to pass.

src/components/page/SideMenu/SideMenu.tsx
```diff
@@ -0,0 +1,75 @@
...
+  ADMIN_NAVIGATION_MODULE_OBJECT,
+  CLIENT_NAVIGATION_MODULE_OBJECT,
+  ENABLE_ADMIN_MODULE,
+  ENABLE_CLIENT_RECORDS_MODULE,
+  REPORT_NAVIGATION_MODULE_OBJECT,
+} from '../../../constants';
+import SubMenu, { NavCollapseObj, NavObj, SubMenuProps } from '../SubMenu/SubMenu';
+import './SideMenu.css';
+
+export interface NavModuleObj {
```
p-netm: forgot to add docstrings?

src/components/page/SideMenu/SideMenu.tsx
```diff
@@ -0,0 +1,75 @@
...
+            {ENABLE_CLIENT_RECORDS_MODULE && <SubMenu {...clientSubMenuProps} />}
+            {ENABLE_REPORT_MODULE && <SubMenu {...reportSubMenuProps} />}
+            {ENABLE_ADMIN_MODULE && <SubMenu {...adminSubMenuProps} />}
+          </Col>
+        </Row>
+      </div>
+    );
+  }
+
+  private setExpandToggle = (sideMenuIdentifier: string): void => {
```
p-netm: forgot to add docstrings?

src/components/page/SideMenu/SideMenu.tsx
```diff
@@ -0,0 +1,75 @@
...
+import SubMenu, { NavCollapseObj, NavObj, SubMenuProps } from '../SubMenu/SubMenu';
+import './SideMenu.css';
+
+export interface NavModuleObj {
+  identifier: string;
+  parentNav: NavCollapseObj;
+  childNavs: NavObj[];
+}
+
+export interface SideMenuState {
```
p-netm: seems like you will always ever have a single string in the expanded nav, if that is the case, do we really want it to be an array

src/components/page/SideMenu/SideMenu.tsx
p-netm: can this function be renamed more intuitively, this might just be me but using set and toggle together is not that intuitive

src/components/page/SubMenu/SubMenu.tsx
redoy11: Well, in my context, isCollapseMenuActive and isExpand represent two different things. isCollapseMenuActive is used to highlight the collapse menu which has its child nav selected. On the other hand, isExpand is used to represent which collapse menu is expanded. There can be cases like the two variables point to two different collapse menus. On such example can be the admin collapse menu is expanded but we are in the household page, so clients collapse menu need to be highlighted.

src/components/page/SubMenu/SubMenu.tsx
redoy11: I will try to rename this variable. The purpose of this flag is to highlight the collapse menu when the current location matches any of the child nav URLs.

src/components/page/SubMenu/SubMenu.tsx
redoy11: The collapse side nav needs to be highlighted. The only way I could think of achieving that is using the withRouter and matching the location property with the child nav URLs. Since I do not know of any event that occurs when the location changes of the router, I used the above function each time when the component renders.

src/constants.tsx
```diff
@@ -1,9 +1,172 @@
...
+};
+
+// nav module objects
+export const CLIENT_NAVIGATION_MODULE_OBJECT: NavModuleObj = {
+  childNavs: [
+    ENABLE_CLIENT_PAGE && CLIENT_PAGE_NAVIGATION_OBJECT,
+    ENABLE_HOUSEHOLD_PAGE && HOUSEHOLD_PAGE_NAVIGATION_OBJECT,
+    ENABLE_ANC_PAGE && ANC_PAGE_NAVIGATION_OBJECT,
+    ENABLE_CHILD_PAGE && CHILD_PAGE_NAVIGATION_OBJECT,
+  ].filter<NavObj>(
```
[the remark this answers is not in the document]
redoy11: I was having a hard time on this. How do you think I should filter the booleans from the NavObj?

src/constants.tsx
```diff
@@ -1,9 +1,172 @@
...
+export const ADMIN_COLLAPSE_NAVIGATION_LABEL = 'Admin';
+export type ADMIN_COLLAPSE_NAVIGATION_LABEL = typeof ADMIN_COLLAPSE_NAVIGATION_LABEL;
+
+// icons
+export const CLIENT_COLLAPSE_NAVIGATION_ICON: IconProp = ['far', 'user'];
+export const REPORT_COLLAPSE_NAVIGATION_ICON: IconProp = ['fas', 'chart-line'];
+export const ADMIN_COLLAPSE_NAVIGATION_ICON: IconProp = ['fas', 'cog'];
+
+// nested labels
+export const CLIENT_PAGE_NAVIGATION_LABEL = 'All Clients';
```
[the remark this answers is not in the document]
redoy11: I will rename it.

src/components/page/SideMenu/SideMenu.tsx
redoy11: Sorry. Missed it. I will add it now.

src/components/page/SideMenu/SideMenu.tsx
redoy11: I think in the future, if we need to store multiple values, this can be helpful. That's why kept it as an array.

src/components/page/SubMenu/tests/SubMenu.test.tsx
redoy11: Well, this happens because all the nav link items point to the same link i.e /404. If I provide different links to different nav link items, this will be resolved. 
Then, I can also add a test as you have mentioned. Hope this is okay.

src/components/page/Header/tests/Header.test.tsx
```diff
@@ -42,7 +42,7 @@ describe('components/page/Header', () => {
         <HeaderComponent {...props} />
       </Router>
     );
-    expect(toJson(wrapper)).toMatchSnapshot();
+    expect(toJson(wrapper.find('Header'))).toMatchSnapshot();
```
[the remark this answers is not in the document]
redoy11: Changed it to the navbar of the Header component

src/components/page/SubMenu/tests/SubMenu.test.tsx
p-netm: ok, but this is still a bug to a user though , so we might need to figure out how to fix this.

src/constants.tsx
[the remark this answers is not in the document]
p-netm: Apparently this is how you have to flter if you want typescript to infer the type of the data returned from the callback.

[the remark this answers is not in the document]
p-netm: :1st_place_medal:

src/App/App.tsx
```diff
@@ -11,22 +12,28 @@ import { providers } from '../configs/settings';
...
 import ConnectedHeader from '../containers/ConnectedHeader';
 
+import SideMenu from '../components/page/SideMenu';
 import { CLIENT_URL } from '../constants';
 import ConnectedClientList from '../containers/Clients/List';
 import Home from '../containers/pages/Home/Home';
 import { oAuthUserInfoGetter } from '../helpers/utils';
 import './App.css';
 
 library.add(faUser);
```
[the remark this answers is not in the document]
p-netm: seen this, will modify accordingly

p-netm:
only pending change on this:
- styling for the login/logout dropdown : ticket number #61

PR #53, opened 2019-09-30
Oath page login page

FrankApiyo:
This PR adds a custom Oath login component and also makes the Header component render null when the current href includes a string. It fixes: #54


FrankApiyo:
I don't know if this is something that should be merged to master because it is very specific to vietnam

moshthepitt:
@FrankApiyo most probably not.  What you need to do is arrange all the Vietnam-specific code to all be merged into one branch.  Once we have it all in one branch, we will have an easier time figuring out what to do.  But we are probably going to use packages.

cc: @cKellyDesign @p-netm

src/components/page/Header/Header.tsx
```diff
@@ -54,6 +54,9 @@ export class HeaderComponent extends React.Component<HeaderProps, State> {
   }
 
   public render() {
+    if (window.location.href.includes('login')) {
```
moshthepitt: Why do we need this hacky-looking thing?

KipSigei:
Just a heads up: I have merged this into `miecd_vietnam` branch cc @FrankApiyo

PR #49, opened 2019-09-19
Coverage reporting

p-netm:
Enables coverage reporting to coveralls

PR #55, opened 2019-10-01
Coverage reporting fix

p-netm:
Attempt at raising branch coverage


p-netm:
any further work on this will be dependent on #49, however i fear that merging more code into master without closing #49 will only add to the load of work needed for this Pr, however i must also mention that *branch* coverage is already above 80%, and i feel its not a bad lower threshold

Issue #65, opened 2019-10-22
Increase test coverage for branch `miecd_vietnam_branch`


FrankApiyo:
-[ ] Write more tests to get a better code coverage for  branch `miecd_vietnam_branch`

FrankApiyo:
**Blocker here: **When rendering components that contain `<Hicharts/>` using `enzyme's` `mount`, an error is thrown that looks something like this:

``` Highcharts error #13: www.highcharts.com/errors/13

      30 |   }
      31 |   public componentDidMount() {
    > 32 |     const chart = Highcharts.chart('chart-wrapper', {
         |                              ^
      33 |       chart: {
      34 |         type: 'line',
      35 |         width: 0.8 * window.innerWidth,

      at c.Chart.Error (node_modules/highcharts/highcharts.src.js:463:31)
      at Object.call [as fireEvent] (node_modules/highcharts/highcharts.src.js:2408:17)
      at Object.fireEvent [as error] (node_modules/highcharts/highcharts.src.js:471:17)
      at c.Chart.error [as getContainer] (node_modules/highcharts/highcharts.src.js:24221:21)
      at c.Chart.getContainer [as firstRender] (node_modules/highcharts/highcharts.src.js:25116:17)
      at c.Chart.firstRender (node_modules/highcharts/highcharts.src.js:23572:21)
      at call (node_modules/highcharts/highcharts.src.js:2408:17)
      at c.Chart.fireEvent [as init] (node_modules/highcharts/highcharts.src.js:23446:17)
      at c.Chart.init [as getArgs] (node_modules/highcharts/highcharts.src.js:23420:17)
      at new apply (node_modules/highcharts/highcharts.src.js:23359:13)
      at Object.c.chart (node_modules/highcharts/highcharts.src.js:23393:20)
      at MotherWeightChart.chart [as componentDidMount] (src/components/MotherWeightChart/index.tsx:32:30)
      at commitLifeCycles (node_modules/react-dom/cjs/react-dom.development.js:17334:22)
      at commitAllLifeCycles (node_modules/react-dom/cjs/react-dom.development.js:18736:7)
      at HTMLUnknownElement.callCallback (node_modules/react-dom/cjs/react-dom.development.js:149:14)
      at Object.invokeGuardedCallbackDev (node_modules/react-dom/cjs/react-dom.development.js:199:16)
      at invokeGuardedCallback (node_modules/react-dom/cjs/react-dom.development.js:256:31)
[...]

Issue #64, opened 2019-10-22
Further cleanup on the `miecd_vietnma_branch`


FrankApiyo:
- [ ] remove warnings logged by the test runner
- [ ] Remove warnings logged on the console while running the app

PR #69, opened 2019-10-23
Increase test coverage for the miecd_vietnam_branch

FrankApiyo:
Fixes: #65


.env.sample
```diff
@@ -61,7 +61,7 @@ REACT_APP_ONADATA_ACCESS_TOKEN_URL=https://api.ona.io/o/token/
 REACT_APP_ONADATA_AUTHORIZATION_URL=https://api.ona.io/o/authorize/
 
 REACT_APP_ONADATA_USER_URL=https://api.ona.io/api/v1/user.json
-REACT_APP_SUPERSET_ANALYSIS_ENDPOINT=https://discover.ona.io/superset/dashboard/25/?standalone=true
+# REACT_APP_SUPERSET_ANALYSIS_ENDPOINT=https://discover.ona.io/superset/dashboard/25/?standalone=true
```
moshthepitt: Why are you commenting this out?

FrankApiyo: this url has been moved to the constants file

FrankApiyo: So this line is commented out because we don't get this url from the `.env` file anymore, it's gotten from the constants file

moshthepitt: why dont you delete it then? @FrankApiyo

FrankApiyo: @moshthepitt. done removed it.

moshthepitt:
We need to fix the warnings logged to the console.

FrankApiyo:
There's a separate issue for this here: https://github.com/OpenSRP/opensrp-web/issues/64. I thought. I'd do that as part of a separate task

moshthepitt:
ok, approved

PR #84, opened 2019-10-29
Addition of Household Register and Profile page

### LINKED ISSUE #33: Create Households Register Table
This is a subset of Issue #22. This involves creating only the household register table in households register view.

- [ ] Create a household register view under the containers.

- [ ] Create a navigation item on the header that directs to that view.

- [ ] Use the household reducer and client's endpoint to fetch the household list.

- [ ] Since the client's interface is used in household reducer, keep blanks to non-matching columns and values for matching columns.

### LINKED ISSUE #67: Create a general search component
The general search component will be used in #21 and other list views.
The following needs to be done to close this issue:

- [ ] Create a separate component named SearchBox

- [ ] props should include a function that callbacks when a search occurs

- [ ] props should include a property to provide a label to the search box

- [ ] the search on the component will occur when the user will press enter key

### LINKED ISSUE #77: Create a single client reducer for the profile view pages.
The reducer should be used for #22.
To close the issue, the following things are required:

- [ ] Create a store state that includes a client object, events object, and members object. The client object is the one that is viewed as a profile. The events and members objects hold the events and members related to that client.
-[] Create actions to add the client object, events object and members object.

- [ ] Create selectors to fetch the state objects.

### LINKED ISSUE #79: Household Basic Information View
The view will be required in #22. To complete the issues the following things need to be done:

- [ ] Create a household profile view page

- [ ] Create a four-column table for the view. The first and third columns are for labels; the others are for respective information. 

- [ ]  Fetch and save related events data and client data from the server to the store.

- [ ] View the table fields from the following data fields: 
       - HH ID Number  _from_ client.BaseEntityId
       - Phone _from_ client.attributes.phoneNumber
       - Family Name _from_ client.lastName
       - Provider _from_ event.providerId
       - Head of Household _from_ client.firstName
       - Register date _from_ client.dateCreated

BasicInformation

### LINKED ISSUE #80: Household Member List view
The view will be required #22 . To complete the issue, the following things need to be done:

- [ ] Create a member list table.

- [ ] Use the _client/searchByCriteria_ endpoint to pull members data and save to redux store

- [ ] Use the client reducer #77 to fetch the member list

- [ ] The member list data mappings are as follows:
      - client.firstName  -> First Name 
      - client.lastName -> Last Name
      - client.attributes.age_year_part -> Age
      - client.attributes.registration_status -> Register Status
memberList

### LINKED ISSUE #82: Navigation to Household Profile Page
- [ ] Create a navigation path to a Household profile page in the App 

- [ ] Use URL _household/profile/:id/_ for Household Path where id will be the baseEntityId of the client Household

### LINKED ISSUE #83: Add the general search to Household List
#21 contains a general search box for searching on the household list. To complete the issue, the following list needs to be completed:

- [ ] Add the search component #67  to Household Register page

- [ ] Add the searchText property to the store state.

- [ ] Create action creators and selectors needed to manipulate the new property

- [ ] Modify reducer based on the new action creators.

redoy11:
This PR closes #33, closes #67, closes #77, closes #79, closes #80, closes #82, closes #83. 
The following things are present in the PR:

-  Addition of Household List page and Household profile page

-  Addition of a new reducer for the Household profile page

-  Addition of a new search box component

-  Addition of the search feature on the Household List page


src/App/App.tsx
```diff
@@ -46,7 +48,18 @@ class App extends Component {
...
+              <ConnectedPrivateRoute
+                disableLoginProtection={DISABLE_LOGIN_PROTECTION}
+                exact={true}
+                path={HOUSEHOLD_URL}
+                component={ConnectedHouseholdList}
+              />
+              <ConnectedPrivateRoute
+                disableLoginProtection={DISABLE_LOGIN_PROTECTION}
+                exact={true}
+                path="/household/profile/:id"
```
p-netm: Lets define this in constants

src/components/page/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/** default state for searchBox */
+const defaultSearchBoxState = {
+  searchText: '',
+};
+
+class SearchBox extends React.Component<SearchBoxProps, SearchBoxState> {
+  public componentDidMount() {
+    this.setState(defaultSearchBoxState);
```
p-netm: I am just curious why this isn't in a constructor

src/components/page/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+  public render() {
+    return (
+      <InputGroup>
+        <InputGroupAddon addonType="prepend">
+          <InputGroupText>
+            <FontAwesomeIcon icon={['fas', 'search']} />
+          </InputGroupText>
+        </InputGroupAddon>
+        <Input placeholder="search" onChange={this.handleChange} onKeyPress={this.search} />
```
p-netm: Am trying to think about this, might be wrong though. So i think onchange for input happens when the input loses focus. From this code is it possible that by the time am clicking enter the onchange will still not have run since input is still in focus and thus the call back will be called with an empty string

src/components/page/SearchBox/SearchBox.tsx
```diff
@@ -0,0 +1,50 @@
...
+  private search = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    const { searchCallBack } = this.props;
+    const { searchText } = this.state;
+    if (event.key === 'Enter') {
+      searchCallBack(searchText);
+    }
+  };
+}
+
+export default SearchBox;
```
p-netm: Nice work on this

src/components/page/SearchBox/SearchBox.tsx
redoy11: Well, I thought the onchange event happens whenever the input value changes, whether it is due to a key press or a copy paste or any action. I may be wrong again but I think it has no relationship with the input focus. Besides, I used the component on Household List search and I did not face the issue with the empty string.

src/components/page/SearchBox/SearchBox.tsx
redoy11: Thank you.

src/components/page/SearchBox/SearchBox.tsx
p-netm: ok

src/store/ducks/clients.ts
```diff
@@ -216,3 +263,19 @@ export function getHouseholdsArray(state: Partial<Store>): Household[] {
...
+  return (state as any)[reducerName].totalRecords;
+}
+
+/** returns the current navigation page
+ * @param {Partial<Store>} state - the redux store
+ * @return { number } - navigation page value from the store
+ */
+export function getNavigationPage(state: Partial<Store>): number {
+  return (state as any)[reducerName].navigationPage;
+}
```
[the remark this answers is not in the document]
p-netm: @redoy11 Would this work
```
.
├── ...
├── store
│   ├── households             
│   ├── clients # records for: currently: (all clients, members) later + : (child, ANC Clients)              
│   ├── events            
```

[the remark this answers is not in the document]
redoy11: I think this can work. But do we really need to separate the households? Households are a type of client. What about we keep only the clients and events. And add households to clients. Also, one other thing, on the list pages, we have things like searchText and pagination number that does not fall into any of the two categories. Do you think it will be better to add a different reducer like common? Will something like below will be okay?
`
├── ...
├── store
│   ├── common             
│   ├── clients # records for: currently: (all households, clients, members) later + : (child, ANC Clients)              
│   ├── events
│   ├── location 
`

src/components/page/SearchBox/SearchBox.tsx
FrankApiyo: I think `onChange` may have a relationship with input focus: https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_onchange

FrankApiyo: @redoy11 you could try to use `oninput` instead

src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+    const response = await clientService.list(params);
+    removeHouseholdsActionCreator();
+    fetchHouseholdsActionCreator(response.clients);
+    setTotalRecordsActionCreator(response.total);
+  }
+}
+
+export { HouseholdList };
+/** Maybe define default props */
+/** connect the component to the store */
```
FrankApiyo: Maybe These comments should be removed

src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  private async getHouseholdList(searchText: string) {
+    const {
+      fetchHouseholdsActionCreator,
+      setTotalRecordsActionCreator,
+      removeHouseholdsActionCreator,
+      opensrpService,
+    } = this.props;
+    const params = {
+      clientType: 'ec_household',
```
FrankApiyo: `ec_household` should have been gotten from a constant file.

src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+/** connect the component to the store */
+
+/** Interface to describe props from mapStateToProps */
+interface DispatchedStateProps {
+  householdsArray: Household[];
+  totalRecordsCount: number;
+}
+
+/** Map props to state  */
+const mapStateToProps = (state: Partial<Store>): DispatchedStateProps => {
```
FrankApiyo: why don't we use something like `Partial<HouseholdListProps>` here instead of `DispatchedStateProps`... T I feel like it's a really bad name for that interface, what do you think about this.

src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+      fetchEventsActionCreator,
+      match,
+    } = this.props;
+    const householdId = match.params.id || '';
+    const params = { identifier: householdId };
+    const clientService = new OpenSRPService(`${OPENSRP_CLIENT_ENDPOINT}`);
+    const clientResponse = await clientService.list(params);
+    if (clientResponse[0]) {
+      fetchClientActionCreator(clientResponse[0]);
+      const eventService = new OpenSRPService('event/search');
```
FrankApiyo: `'event/search'` string should also go to some kind of constant just like `OPENSRP_CLIENT_ENDPOINT` that will let us know what endpoint it is.

FrankApiyo: perhaps `OPENSRP_EVENTS_ENDPOINT`?

src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+/** Interface to describe props from mapStateToProps */
+interface DispatchedStateProps {
+  household: Household;
+  events: Event[];
+  members: Household[];
+}
+
+/** Map props to state  */
+const mapStateToProps = (state: Partial<Store>): DispatchedStateProps => {
```
FrankApiyo: how about `Partial<Props>` instead of creating a whole new interface

src/store/ducks/client.ts
```diff
@@ -0,0 +1,192 @@
...
+ * @param {Event} events - events to add to store
+ * @return {FetchEventsAction} - an action to add events to redux store
+ */
+export const fetchEvents = (eventsList: Event[]): FetchEventsAction => ({
+  eventsById: keyBy(eventsList, (event: Event) => event.baseEntityId),
+  type: EVENTS_FETCHED,
+});
+
+/** Fetch members action creator
+ * @param {Event} members - members to add to store
```
FrankApiyo: Perhaps we should change this type `Event` to `Client`

src/store/ducks/client.ts
```diff
@@ -0,0 +1,192 @@
...
+  return (state as any)[reducerName].membersById;
+}
+
+/** returns the members of the client as array
+ * @param {Partial<Store>} state - the redux store
+ * @return { Client } - members array
+ */
+export function getMembersArray(state: Partial<Store>): Client[] {
+  return values((state as any)[reducerName].membersById);
+}
```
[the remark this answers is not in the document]
FrankApiyo: Maybe `Event`s at least, should have their own reducer module?

src/containers/Households/List/index.tsx
redoy11: I will move this to the constant file.

src/containers/Households/List/index.tsx
redoy11: I think that makes more sense.

src/store/ducks/clients.ts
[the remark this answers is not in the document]
p-netm: so the initial goal was to have a redux store structure that rudimentarily looks like this:
```
.
├── ...
├── store
│   ├── households             
│   ├── child          
│   ├── ANC
│   ├── members
│   ├── events
```
the reasoning was that while the server does not do a great of a job in normalizing this data, does not mean we can't do things a little better on the web, 

the aspect driving how we separate the clients is based on how we semantically idetify the type of clients.
- for instance while households is technically a client, we understand a household to be different from other types of clients like say a `child`.

However we had a discussion and we found the above approach to have shortcomings given the underlying business logic.

__so why are we not putting all clients in a single store__
- code maintainability: makes it much easier to modify a smaller piece of a larger whole, PS: where the smaller piece is possible and logical like in this case(IMO).
    - households is a smaller piece of the clients ducks.
- semantics: would give meaning to the fetched data long before it goes through any preprocessing for display
- Portability: from conversation with Hildah seems like we might have variants of this web app that don't need certain clients data say the  households. having the ducks module more normalized would allow for much easier portability, 

__Proposed Structure__:
To split households out of clients ducks modules, unfortunalely we cannot do the same for other types of clients yet.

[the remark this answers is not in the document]
p-netm: If something does not make sense, just flag it and we will converse over that.

src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    } = this.props;
+    const params = {
+      clientType: 'ec_household',
+      pageNumber: '0',
+      pageSize: PAGINATION_SIZE,
+      searchText,
+    };
+    const clientService = new opensrpService(`${OPENSRP_HOUSEHOLD_ENDPOINT}`);
+    const response = await clientService.list(params);
+    removeHouseholdsActionCreator();
```
p-netm: So at the moment: the architecture i would propose is that we can have a state variable that tracks the search text, e.g you could track the text as some form of filter,.

`onChange` `event` on the `searchBox` would do 2 things: 
- modify the state variable tracking related filters and
-  also initiate an api call that only adds the fetched data to store; note we dont have to use the `removeAction`

when displaying the data we then just need to preprocess the data got from the store using state tracked filters to get the slice of data that should be shown on the table. Here, also note that the full data got from the store will still be accessible from within the component as a prop. and the filtered data will be in a state variable.

You did raise one interesting point on adding a flag to component that has information on status of api calls so as to know if we received any data from a resolved api call. i think this could be useful even with the above mentioned architecture, I however have not thought about its implementation and thus do not have any ideas on how it would used.

p-netm: @redoy11

redoy11: @p-netm  Well, I think I have understood now. But at the same time, I think preprocessing the data from the store using state tracked filters will be challenging and also, honestly, it looks unnecessary to me. We actually do not need any preprocessing since the server is sending the exact slice that needs to be shown in the table. We just need to map the interesting incoming attributes with the table attributes since the server is handling all the filters. If we append the API response to the store state clients variable and filter data from the stored state to the component state, we are just introducing an unnecessary preprocessing. For the same and exact slice that will be returned by the API, we will be mixing it, preprocessing it and retrieving it from the store.  Also, if we save it directly to the component state, we will be making the store state of clients variable redundant. Overall, I may be wrong, it seems to add unnecessary complexity without any actual benefit.
Regarding the flag, I think sth that tracks the first API response since we will be showing the loading component only when the first slice of data has not appeared. But it will be helpful if we can make sth generic regarding whenever we fetch sth from the server. Remember the time I told you that there is a brief interval of time lapse between the time I search sth and the reflection of it on the table. I think a component like loading during that time will ease the user experience. In addition, I think we also require it in the long run since the time will increase as the size of DB increases.

src/store/ducks/clients.ts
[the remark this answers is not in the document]
redoy11: @p-netm One small question though, will we be extending the client interface to the household or use the same interface as the client?

src/containers/Households/List/index.tsx
p-netm: >  state tracked filters will be challenging

could be, but does not have to be, we could possibly discuss how this could be done as easily as possible.

p-netm: >  Overall, I may be wrong, it seems to add unnecessary complexity without any actual benefit

For one i think you the complexity you mention is overstated,.. as for the benefit, well i can simply put it as responsiveness and thus UX, see,.. if we persist data from api call A and then later on need the same data in a different view say view B; it means that we can render a preview of the data albeit "possibly" stale, before the api call B from within view B resolves. When call B resolves it rerenders the view with the fresh data and all the while the user will not have to wait for any load time to start interacting with the data.

On the other hand if we were to keep emptying the store and repopulating it with new data on each api fetch, then the clients ducks module becomes an unnecessary overhead, because we could simply do this from within the compnent itself. i.e like in the classic way, just fetch data and render it from within where its needed for display.

src/store/ducks/clients.ts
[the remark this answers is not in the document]
p-netm: i think we can factor out the interface into a separate file and import it from within the ducks modules that need it. This is of course assuming that the interface is wholly similar for both households and clients.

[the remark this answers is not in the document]
p-netm: @redoy11

src/containers/Households/List/index.tsx
p-netm: i think `DispatchedStateProps` si more explicit about how the returned object should look like,.. IMHO typescript works best when we are very explicit and concise about our typings and if asked , i would say `DispatchedStateProps` works better than `Partial<Props>`

src/containers/Households/List/index.tsx
p-netm: > Regarding the flag, I think sth that tracks the first API response since we will be showing the loading component only when the first slice of data has not appeared

makes sense, i am researching a few possible ways we can do this, but for the moment i think your proposed approach will suffice

> I think a component like loading during that time will ease the user experience. In addition

I agree, we need to add some more loading components that best fit this useCase.

redoy11: > it means that we can render a preview of the data albeit "possibly" stale, before the api call B from within view B resolves. When call B resolves it rerenders the view with the fresh data and all the while the user will not have to wait for any load time to start interacting with the data.

Since we have pagination, a user can jump from pages. So if a user surf page 1, then page 10 and then page 4, how do you propose we create the preview of page 4 from the records of only page1 and page 10?

p-netm: I may have misunderstood the scenario here, but what i imagine is that, if you originally have 10 pages worth of records, if you search and say the response is a single record then the component will rerender with the table having only the one record and therefore pagination will have a single page,.. at this point we still have all the records that were initially fetched in the store plus any that was fetched from calls resulting from the search.
Where we need to track the the component's current pagination page we could still do that from within the state so that we can always show the correct page in between rerenders, this also goes for things like the scroll position if  say we have more records in the table than those that can fit the available viewport height.

redoy11: > > state tracked filters will be challenging
> 
> could be, but does not have to be, we could possibly discuss how this could be done as easily as possible.

Well, I have an idea. It can be something like we always filter the store clients based on the filter criteria stored in the component state before render. But at the same time, we will need to create the same filter criteria in the redux store. The onChange or onSearch handler will always set the filter criteria of the component only. And on the componentDidUpdate method, we can check whether the filter criteria of the component state and redux store matches. If it does not match, we update the filter criteria of the store with the component state and fetch new data from the server to the redux store. if it matches, we do not take any action. How does this sound?

redoy11: > if you search and say the response is a single record then the component will rerender with the table having only the one record and therefore pagination will have a single page

Well, I thought we are setting the pagination number of pages from the total records. The API returns the total number of records present in the server. So I thought we are setting the pagination pages from there. Next, the API also has a pageNumber parameter and so say, if we want records from page 10, we can set the pageNumber to 10 and retrieve only the records that will actually be present on page 10. So if a user surfs directly from page 1 to page 10, skipping the interval pages, the store will only have records from page 1 and page 10. Now if the user surfs on page 4, how will we build the preview from the records of only page 1 and page 10 or how we will handle the pagination while building the previews. I hope I am able to clear the scene.

src/components/page/SearchBox/SearchBox.tsx
redoy11: @FrankApiyo I think the react onChange works just like onInput. You can see the following link for details -  https://stackoverflow.com/questions/38256332/in-react-whats-the-difference-between-onchange-and-oninput

src/App/App.tsx
cKellyDesign: @p-netm @moshthepitt - Should the path params remain?
```
path={`${HOUSEHOLD_PROFILE_URL}/:id`}
```

Or should the whole thing go?
```
path={HOUSEHOLD_PROFILE_URL}
```

src/components/page/SearchBox/SearchBox.tsx
cKellyDesign: This should go in the constructer

src/App/App.tsx
p-netm: @cKellyDesign  i would go with the first over the latter

src/store/ducks/households.ts
```diff
@@ -0,0 +1,145 @@
...
+import { get, keyBy, keys, values } from 'lodash';
+import {Client} from './clients';
+import SeamlessImmutable from 'seamless-immutable';
+
+/** The reducer name */
+export const reducerName = 'households';
+
+export const clientType = 'ec_household';
+
+export const searchPlaceholder = 'Search Household';
```
p-netm: This does not feel like it belongs here, i feel as if its best defined where its being used,...

src/store/ducks/households.ts
```diff
@@ -0,0 +1,145 @@
...
+
+/** Interface for household object same as client */
+export type Household = Client;
```
p-netm: maybe we can differentiate the interfaces based on the clienttype  property, from line 9

src/store/ducks/members.ts
```diff
@@ -0,0 +1,76 @@
+import { AnyAction, Store } from 'redux';
```
p-netm: I think the proposed ducks module structure does not have members in it.

src/components/page/SearchBox/tests/SearchBox.test.tsx
```diff
@@ -1,16 +1,23 @@
...
-    let props = {
-        searchCallBack: jest.fn(),
-        placeholder: "Search Household"
-    }
     const component = shallow(<SearchBox {...props} />);
     const wrapper = component.find("#custom-search-input");
     expect(wrapper.length).toBe(1);
   });
+
+  it('has proptypes and checks the prop values', ()=> {
```
p-netm: what functionality would this test be checking?

Its really nice to start seeing tests on this repo.

prodhan29: checking the props values passed by household,  whether it passed the exact value or not.

p-netm: I think this test is surplus, one question on this would be what conditions do you see would result in this test failing.
Another issue i have with this test is that it seems to exercise react's core functionality which is not really what we should be testing.

src/store/ducks/members.ts
redoy11: I thought we are following the structure of the last comment (https://github.com/OpenSRP/opensrp-web/pull/84#discussion_r345224649).  This had members as a separate ducks module but do you want us to refractor this to clients?

src/store/ducks/households.ts
redoy11: Well, are you referring to the following line
`export const clientType = 'ec_household';`
If so, the clientType referred here is something that is used only on the API URL params, not a property of the client object itself.

src/store/ducks/members.ts
p-netm: > However we had a discussion and we found the above approach to have shortcomings given the underlying business logic.

@prodhan29 @redoy11 This was the proposed approach [comment](https://github.com/OpenSRP/opensrp-web/pull/84#discussion_r341628365.)

src/store/ducks/households.ts
p-netm: Oouh!! ok.

src/store/ducks/households.ts
prodhan29: moved them to the global  constant file.

> src/constants.tsx

src/components/page/SearchBox/tests/SearchBox.test.tsx
prodhan29: acknowledged.

src/App/App.tsx
prodhan29: Refactor the url path.

src/components/page/SearchBox/SearchBox.tsx
prodhan29: it is now in the constructor.

src/containers/Households/List/index.tsx
prodhan29: moved this to the constant file

PR #88, opened 2019-10-30
Add paginator to log face

FrankApiyo:
Fixes: #86

### DIFF
--- a/src/containers/LogFace/tests/index.test.tsx
+++ b/src/containers/LogFace/tests/index.test.tsx
@@ -52,23 +52,6 @@ describe('components/ConnectedHeader', () => {
     wrapper.unmount();
   });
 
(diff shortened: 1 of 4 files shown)


src/containers/LogFace/tests/index.test.tsx
```diff
@@ -52,23 +52,6 @@ describe('components/ConnectedHeader', () => {
...
-      </Provider>
-    );
-
-    expect(wrapper.find('#next').length).toBe(1);
-    expect(wrapper.find('#previous').length).toBe(0);
-    wrapper.find('#next').simulate('click');
-    expect(wrapper.find('#previous').length).toBe(1);
-    wrapper.unmount();
-  });
-
```
FrankApiyo: This test has been removed because the old paginator is no-longer there

PR #89, opened 2019-10-30
Add periodic fetching of missing sms events

FrankApiyo:
This PR accomplishes the following: 

1. Add periodic fetching of missing **sms events** from **supserset** using the `getform` option
2. Move all slice IDs and superset URLs to the `.env` file.
3. Replaces the **slice ID** used in the `LogFace` component with a prop.


src/containers/LogFace/index.tsx
```diff
@@ -97,12 +100,41 @@ export class LogFace extends React.Component<PropsInterface, State> {
     };
   }
 
+  public sortFunction = (firstE1: SmsData, secondE1: SmsData): number => {
```
KipSigei: Can we make this a util function so that we can reuse on other components that might need it

src/components/page/SideMenu/SubMenu/index.css
```diff
@@ -43,3 +46,11 @@
   padding-right: initial;
   padding-left: 33px;
 }
+
+.active-collapse-menu path {
+  fill: #0192d4;
+}
+
+path {
```
KipSigei: This style will apply on any svg in any mounted component, can we add a class so that this doesn't change on other icons and make it specific to the sidemenu only?

src/superset-connector.d.ts
```diff
@@ -24,6 +24,7 @@ declare module '@onaio/superset-connector' {
     authZ: any;
```
moshthepitt: This file needs to be deleted.  Superset Connector has its own type defs: https://github.com/onaio/js-tools/tree/master/packages/SupersetConnector/dist/types

src/components/page/SideMenu/SubMenu/index.css
FrankApiyo: cool

src/containers/LogFace/index.tsx
```diff
@@ -105,6 +110,28 @@ export class LogFace extends React.Component<PropsInterface, State> {
...
         fetchSmsDataActionCreator(result);
       });
+    } else {
+      const intervalId: NodeJS.Timeout = setInterval(() => {
+        const smsDataInDescendingOrderByEventId: SmsData[] = this.props.smsData.sort(sortFunction);
+
+        // pick the lartgest ID if this smsDataInDescendingOrderByEventId list is not empty
+        if (smsDataInDescendingOrderByEventId.length) {
+          const largestEventID: string = smsDataInDescendingOrderByEventId[0].event_id;
+          const supersetParams = superset.getFormData(2000, [
```
KipSigei: lets move this rowLimit to an env var

FrankApiyo:
What's left: The SLICE IDs need to be moved to the `.env` file.

src/containers/LogFace/index.tsx
```diff
@@ -105,6 +110,28 @@ export class LogFace extends React.Component<PropsInterface, State> {
...
+            { comparator: largestEventID, operator: '>', subject: 'event_id' },
+          ]);
+          supersetFetch(SUPERSET_SMS_DATA_SLICE, supersetParams)
```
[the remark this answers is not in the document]
FrankApiyo: done

src/containers/LogFace/index.tsx
FrankApiyo: and done

src/App/Routes.tsx
```diff
@@ -108,12 +108,6 @@ export const Routes = (props: RoutesProps) => {
               />
             )}
           />
-          <ConnectedPrivateRoute
-            disableLoginProtection={false}
-            exact={true}
-            path={CLIENT_URL}
-            component={ConnectedClientList}
-          />
```
FrankApiyo: we do not need/have this URL in MIECD

src/containers/LogFace/index.tsx
```diff
@@ -102,10 +110,31 @@ export class LogFace extends React.Component<PropsInterface, State> {
...
-      supersetFetch(SUPERSET_SMS_DATA_SLICE).then((result: SmsData[]) => {
+      supersetFetch(this.props.sliceId).then((result: SmsData[]) => {
         fetchSmsDataActionCreator(result);
       });
     }
+    const intervalId: NodeJS.Timeout = setInterval(() => {
+      const smsDataInDescendingOrderByEventId: SmsData[] = this.props.smsData.sort(sortFunction);
+
+      // pick the lartgest ID if this smsDataInDescendingOrderByEventId list is not empty
+      if (smsDataInDescendingOrderByEventId.length) {
```
cKellyDesign: Is there ever a case where there are not SmsData in the store at this point?

src/containers/LogFace/index.tsx
```diff
@@ -102,10 +110,31 @@ export class LogFace extends React.Component<PropsInterface, State> {
...
+        const largestEventID: string = smsDataInDescendingOrderByEventId[0].event_id;
+        const supersetParams = superset.getFormData(GET_FORM_DATA_ROW_LIMIT, [
+          { comparator: largestEventID, operator: '>', subject: 'event_id' },
```
cKellyDesign: Let's move `event_id` into the constants to avoid magical strings :)

src/containers/LogFace/index.tsx
```diff
@@ -297,6 +326,12 @@ export class LogFace extends React.Component<PropsInterface, State> {
     );
   }
 
+  public componentWillUnmount() {
+    if (this.state.intervalId) {
+      clearInterval(this.state.intervalId);
+    }
+  }
```
cKellyDesign: Nice

src/App/Routes.tsx
```diff
@@ -155,19 +149,25 @@ export const Routes = (props: RoutesProps) => {
...
+              <ConnectedLogFace header={PREGNANCY} sliceId={SUPERSET_SMS_DATA_SLICE} />
+            )}
           />
           <ConnectedPrivateRoute
             exact={false}
             path={PNC_AND_NBC_LOGFACE_URL}
             // tslint:disable-next-line: jsx-no-lambda
-            component={() => <ConnectedLogFace header={NBC_AND_PNC} />}
+            component={() => (
+              <ConnectedLogFace header={NBC_AND_PNC} sliceId={SUPERSET_SMS_DATA_SLICE} />
```
cKellyDesign: Why pass `SUPERSET_SMS_DATA_SLICE` as a prop rather than importing it directly into the component? Since these are the same component it seems like it would be a bit DRY'er to import in the component rather than here.

src/containers/LogFace/index.tsx
```diff
@@ -33,6 +36,7 @@ import TestReducer, {
   smsDataFetched,
 } from '../../store/ducks/sms_events';
 import './index.css';
+// import { getFormData } from '@onaio/superset-connector/dist/types/utils';
```
cKellyDesign: Let's remove this if it's not going to be used

src/containers/PatientDetails/index.tsx
```diff
@@ -57,9 +58,11 @@ export class PatientDetails extends Component<Props, State> {
   public render() {
     return (
       <div className="patient-details">
-        <Link to={PREGNANCY_LOGFACE_URL} className="back-page">
-          <FontAwesomeIcon icon={BACKPAGE_ICON} size="lg" />
-          <span>{BACK}</span>
+        <Link to="#" className="back-page">
+          <span onClick={this.props.history.goBack}>
```
cKellyDesign: Nice

src/App/Routes.tsx
cKellyDesign: @KipSigei just explained that later different logface instances will use different slices 👍

src/containers/LogFace/index.tsx
FrankApiyo: done

src/containers/LogFace/index.tsx
FrankApiyo: done

PR #101, opened 2019-11-08
Increase branch coverage

### LINKED ISSUE #92: Circle CI Build is failing
Need to raise the branch coverage from ~78 to above the threshold which is 80

p-netm:
Master build is failin since branch coverage is wanting, this Tries to fix that.
closes #92

### DIFF
--- a/src/components/page/SideMenu/SubMenu/tests/index.test.tsx
+++ b/src/components/page/SideMenu/SubMenu/tests/index.test.tsx
@@ -89,7 +89,7 @@ describe('components/page/SubMenu', () => {
     wrapper.unmount();
   });
 
-  it('stimulates click and calls mock function properly', () => {
+  it('simulates click and calls mock function properly', () => {
     const mockCallBack = jest.fn();
     const mock = jest.fn();
     const pathname = '/somePath';
@@ -112,7 +112,7 @@ describe('components/page/SubMenu', () => {
     wrapper.unmount();
   });
 
-  it('stimulates click and calls mock function with correct value', () => {
+  it('simulates click and calls mock function with correct value', () => {
     const mockCallBack = jest.fn();
     const mock = jest.fn();
     const pathname = '/somePath';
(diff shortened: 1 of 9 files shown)


src/configs/env.ts
```diff
@@ -83,32 +83,6 @@ export type OPENSRP_USER_URL = typeof OPENSRP_USER_URL;
...
-export const ONADATA_USER_URL =
-  process.env.REACT_APP_ONADATA_USER_URL || 'https://stage-api.ona.io/api/v1/user.json';
-export type ONADATA_USER_URL = typeof ONADATA_USER_URL;
-
-export const ONADATA_OAUTH_STATE = process.env.REACT_APP_ONADATA_OAUTH_STATE || 'onadata';
-export type ONADATA_OAUTH_STATE = typeof ONADATA_OAUTH_STATE;
-
-export const GISIDA_ONADATA_API_TOKEN = process.env.REACT_APP_GISIDA_ONADATA_API_TOKEN || '';
-export type GISIDA_ONADATA_API_TOKEN = typeof GISIDA_ONADATA_API_TOKEN;
-
```
FrankApiyo: alright.

src/components/page/SideMenu/SubMenu/tests/index.test.tsx
```diff
@@ -111,4 +111,48 @@ describe('components/page/SubMenu', () => {
     expect(mockCallBack.mock.calls.length).toEqual(1);
     wrapper.unmount();
   });
+
+  it('stimulates click and calls mock function with correct value', () => {
```
FrankApiyo: you meant `simulate` here I think.

p-netm:
Decided to revert on the decision to remove onadata oauth, i think there would need for a team discussion if its something that should still remain in the codebase even though IMO its not currently used.

src/components/page/SideMenu/SubMenu/tests/index.test.tsx
p-netm: yeah

src/configs/env.ts
p-netm: Went back on this decision.

FrankApiyo: These code that is not in use should totally be removed.

p-netm: @hmakandi @cKellyDesign Do you think we should keep the oauth settings for onadata as a provider?

cKellyDesign: @p-netm @FrankApiyo - Is OpenSRP able to authenticate with Ona's oAuth provider? If so then it should remain an option, if not then it should be removed.

Issue #107, opened 2019-11-18
QA MIECD Dashboard 4


hmakandi:
**Location Hierarchy**

- [x] Problem: The current locations only work with specific levels. This is because it only accepts one location as an argument. This is a problem when a location is expected to show the constituent children locations, e.g. a Level 1(Country location). Thus there is a need to traverse the authenticate API, get all children locations, and pass them as an array of locations to be filtered from the data set.

- [x] The logface shows all data for all Users. It should show the data for the user that is currently logged in. The higher the location is in the hierarchy the more the children location. This location is based on the user.

- [x] The filters do not work as expected. When selected, they need to work logically by appending an AND kind of functionality

hmakandi:
Confirmed.

PR #112, opened 2019-11-21
Filter data by user location

FrankApiyo:
Fixes: #107

This PR also does the following:

- Rearrange the variables in the `.env.sample` file to get it more organized, easier to locate variables.

- update the gatekeeper package

- deactivate the nutrition module links to compartment and analysis

- remove static user location ID and user a dynamic location ID based on the currently logged in user.

- add ability to log out from the authentication server(OpenSRP) when the user clicks logout

-  Refactor components by moving some reusable functions to the utils

- add more tests.


src/components/VillageData/index.tsx
```diff
@@ -64,7 +63,7 @@ export default class VillageData extends React.Component<Props, State> {
     };
     return (
       <React.Fragment>
-        {this.props.current_level === 3 ? (
+        {this.props.current_level === 3 || this.props.current_level === 4 ? (
```
cKellyDesign: Could this be `this.props.current_level >= 3` instead?

src/components/page/Header/Header.tsx
```diff
@@ -94,6 +101,17 @@ export class HeaderComponent extends React.Component<HeaderProps, State> {
     );
   }
 
+  private handleLogout() {
+    const logoutURL: string = 'https://opensrp-miecd-stage.smartregister.org/opensrp/logout.do';
```
cKellyDesign: Should this go in env?

src/containers/Compartments/index.tsx
```diff
@@ -91,186 +131,110 @@ class Compartments extends React.Component<Props, State> {
...
+      const self: any = this;
+      headers.append(
+        'Authorization',
+        `Bearer ${(session as any).extraData.oAuth2Data.access_token}`
+      );
+      fetch(`${OPENSRP_API_BASE_URL}/security/authenticate`, { headers }).then((user: any) =>
+        user.json().then((res: FlexObject) => {
+          self.props.fetchUserIdActionCreator((res as any).user.attributes._PERSON_UUID);
+        })
+      );
```
cKellyDesign: Do we not have services for this? Like in Reveal?

src/containers/LogFace/index.tsx
```diff
@@ -113,8 +190,56 @@ export class LogFace extends React.Component<PropsInterface, State> {
...
+      const headers: any = new Headers();
+      headers.append(
+        'Authorization',
+        `Bearer ${(session as any).extraData.oAuth2Data.access_token}`
+      );
+      fetch(`${OPENSRP_API_BASE_URL}/security/authenticate`, { headers }).then((user: any) =>
+        user.json().then((res: FlexObject) => {
+          this.props.fetchUserIdActionCreator((res as any).user.attributes._PERSON_UUID);
+        })
+      );
```
cKellyDesign: Same as above, service?

src/helpers/utils.tsx
```diff
@@ -28,11 +41,30 @@ export function oAuthUserInfoGetter(apiResponse: { [key: string]: any }): Sessio
...
-export function headerShouldNotRender() {
+/**
+ * determines weather the header should be rendered.
+ */
+export function headerShouldNotRender(): boolean {
   return RegExp(URLS_TO_HIDE_HEADER.join('|')).test(window.location.pathname);
 }
 
-export function getNumberSuffix(num: number) {
+export function groupBy(data: FlexObject[], field: string) {
```
cKellyDesign: Doc-strings?

src/components/VillageData/index.tsx
FrankApiyo: done!

src/components/page/Header/Header.tsx
FrankApiyo: gone!

src/containers/Compartments/index.tsx
FrankApiyo: Replaced with `SupsersetService`.

src/containers/LogFace/index.tsx
FrankApiyo: Replaced with `SupsersetService`.

.env.sample
```diff
@@ -10,23 +10,15 @@ REACT_APP_GET_FORM_DATA_ROW_LIMIT=2000
...
-REACT_APP_OPENSRP_AUTHORIZATION_URL=https://reveal-stage.smartregister.org/opensrp/oauth/authorize
+REACT_APP_OPENSRP_ACCESS_TOKEN_URL=https://opensrp-miecd-stage.smartregister.org/opensrp/oauth/token
+REACT_APP_OPENSRP_API_BASE_URL=https://opensrp-miecd-stage.smartregister.org/opensrp
+REACT_APP_OPENSRP_AUTHORIZATION_URL=https://opensrp-miecd-stage.smartregister.org/opensrp/oauth/authorize
 REACT_APP_OPENSRP_CLIENT_ID=hunter1
-REACT_APP_OPENSRP_USER_URL=https://reveal-stage.smartregister.org/opensrp/user-details
+REACT_APP_OPENSRP_USER_URL=https://opensrp-miecd-stage.smartregister.org/opensrp/user-details
+REACT_APP_OPENSRP_LOGOUT_URL='https://opensrp-miecd-stage.smartregister.org/opensrp/logout.do'
 REACT_APP_ENABLE_OPENSRP_OAUTH=true
 REACT_APP_REACT_APP_OPENSRP_OAUTH_STATE=opensrp
```
moshthepitt: why does the sample env file need to change?

src/components/page/Header/Header.tsx
```diff
@@ -94,6 +101,17 @@ export class HeaderComponent extends React.Component<HeaderProps, State> {
...
+    const logoutURL: string = OPENSRP_LOGOUT_URL;
+    const logoutWindow: Window | null = window.open(logoutURL);
+    const timer: NodeJS.Timeout = setInterval(() => {
+      if (logoutWindow) {
+        logoutWindow.close();
+      }
+      clearInterval(timer);
+    }, 20);
+  }
+
```
moshthepitt: I think we should remove this logout stuff in favor of https://github.com/onaio/js-tools/pull/109

.env.sample
FrankApiyo: to make the `.env` variables a little more organized. easier to work with and look at.

src/components/page/Header/Header.tsx
FrankApiyo: sure... we can do this now.

src/components/DataCircleCard/index.tsx
```diff
@@ -9,16 +9,14 @@ import {
...
   NBC_AND_PNC_WOMAN,
   NO,
   NO_RISK,
   NUTRITION,
-  NUTRITION_COMPARTMENTS_URL,
   PREGNANCY,
-  PREGNANCY_COMPARTMENTS_URL,
   SMS_FILTER_FUNCTION,
 } from '../../constants';
+import { getModuleLink } from '../../helpers/utils';
```
moshthepitt: Could we add some docstrings in this file?

src/components/DataCircleCard/index.tsx
```diff
@@ -64,23 +42,27 @@ function DataCircleCard({
...
+  userLocationId,
+  permissionLevel,
 }: Props) {
-  // this should be gotten dynamically based on the logged in user
-  // level is a integer ranging from 0 to 3, 0 for province, 1 for
-  // 1 for District, 2 for commune and 3 for village.
-  const level = 1;
-  const locationId = 'eccfe905-0e03-4188-98bc-22f141cccd0e';
-  const permissionLevel = 1;
+  function getLinkToHierarchichalDataTable(riskLevel: HIGH | LOW | NO) {
```
moshthepitt: Why does this function have to be nested this way?  Can it be moved out?

FrankApiyo: should be moved out.

src/components/DataCircleCard/index.tsx
```diff
@@ -64,23 +47,18 @@ function DataCircleCard({
...
     <Card className={`dataCircleCard ${className}`}>
       <CardTitle>{title}</CardTitle>
       <CardBody>
         <ul className="circlesRow">
           <li className="red">
             <Link
-              to={`${getModuleLink(
-                module
-              )}${HIERARCHICAL_DATA_URL}/${module}/${HIGH}/${title}/${level}/down/${locationId}/${permissionLevel}`}
+              to={getLinkToHierarchichalDataTable(HIGH, module, title, permissionLevel, locationId)}
```
cKellyDesign: Nice 👍

src/helpers/utils.tsx
```diff
@@ -28,11 +45,35 @@ export function oAuthUserInfoGetter(apiResponse: { [key: string]: any }): Sessio
...
+  const dataMap: FlexObject = {};
+  list.forEach((listElement: FlexObject) => {
+    if (!dataMap[listElement[field]]) {
+      dataMap[listElement[field]] = {
+        ...listElement,
+      };
+    }
+  });
+  return dataMap;
+}
```
cKellyDesign: What is this returning? Can we have more specific typings than `FlexObject`? Let's also add `@return` to these doc-strigs please 🙏

src/helpers/utils.tsx
```diff
@@ -60,6 +101,118 @@ export const sortFunction = (firstE1: SmsData, secondE1: SmsData): number => {
...
+export function getLocationId(userLocationData: UserLocation[], userUUID: string) {
+  const userDetailObj =
+    userLocationData.length &&
+    userLocationData.find(
+      (userLocationDataItem: UserLocation) => userLocationDataItem.provider_id === userUUID
+    );
+  if (userDetailObj) {
+    return userDetailObj.location_id;
+  }
+}
```
cKellyDesign: Description and `@return` pls

src/helpers/utils.tsx
```diff
@@ -60,6 +101,118 @@ export const sortFunction = (firstE1: SmsData, secondE1: SmsData): number => {
...
+  return locations.find((location: Location) => location.location_id === locationId);
+};
+
+/**
+ * An object representing the filter function and location level for a logged in user.
+ */
+export interface FilterFunctionAndLocationLevel {
+  locationFilterFunction: (smsData: SmsData) => boolean;
+  locationLevel: number;
+}
```
cKellyDesign: Can this be moved to / imported from a `types.ts` file?

src/helpers/utils.tsx
```diff
@@ -60,6 +101,118 @@ export const sortFunction = (firstE1: SmsData, secondE1: SmsData): number => {
...
+
+/**
+ * calculate the filter function and location level for a logged in user
+ * @param userLocationId - the location ID of the logged in user.
+ * @param provinces - a list of locations of level province
+ * @param districts - a list of locations of level district
+ * @param communes - a list of locations of level commune
+ * @param villages - a list of locations of level village
+ */
+export function getFilterFunctionAndLocationLevel(
```
cKellyDesign: Nice description and params, let's add `@returns` as well

src/helpers/utils.tsx
```diff
@@ -60,6 +101,118 @@ export const sortFunction = (firstE1: SmsData, secondE1: SmsData): number => {
...
+    userLocationLevel = 4;
+    locationFilterFunction = (smsData: SmsData): boolean => {
+      return userLocationId === smsData.location_id;
+    };
+  }
+
+  return {
+    locationFilterFunction,
+    locationLevel: userLocationLevel,
+  } as FilterFunctionAndLocationLevel;
```
cKellyDesign: Interesting approach here on returning the filter function like this, I think as long as this is being returned into a clean closure then it shouldn't memory leak. (otherwise the garbage-collector wouldn't re-alocate memory for all the instances of `locationFilterFunction`)

src/store/ducks/locations.ts
```diff
@@ -14,8 +14,26 @@ export interface Location {
...
 }
 
+export interface UserLocation {
+  provider_name: string;
+  provider_contact: string;
+  provider_id: string;
+  location_id: string | any;
+  openmrs_identifier?: string;
+  location_name: string;
+}
```
cKellyDesign: Docstring 🙏

src/store/ducks/locations.ts
```diff
@@ -95,6 +200,27 @@ export function getLocations(state: Partial<Store>): Location[] {
...
+ */
+export function getUserId(state: Partial<Store>): string {
+  return (state as any)[reducerName].userId;
+}
+
+/**Returns the locationId of user
+ */
+export function getUserLocationId(state: Partial<Store>): string {
+  return (state as any)[reducerName].userLocationId;
+}
```
cKellyDesign: Seems pretty intuitive but let's add `@return` here too :)

src/store/ducks/sms_events.ts
```diff
@@ -80,7 +85,7 @@ export type SmsActionTypes =
  */
 export const fetchSms = (smsDataList: SmsData[] = []): FetchSmsAction => {
   return {
-    smsData: keyBy(smsDataList, (smsData: SmsData) => smsData.event_id),
+    smsData: groupBy(smsDataList, 'event_id'),
```
cKellyDesign: Move to `constants.ts`? (cause it's a magic string)

src/helpers/utils.tsx
FrankApiyo: @cKellyDesign I think this way of doing it is elegant. also since every time the `DataCircleCard` component un-mounts, the filter functions are cleared, we will loose all references to them and this will then allow memory allocated to `locationFilterFunction` to be re-allocated.

src/store/ducks/sms_events.ts
FrankApiyo: :+1:  such a constant already exists.

src/helpers/utils.tsx
FrankApiyo: This method is similar to **lodash's keyBy**; so it can receive any key-value pair and return any key-value pair. so... it's hard to get more specific than `FlexObject`

src/helpers/utils.tsx
FrankApiyo: would we have the `types.ts` in the **helpers** directory or in the **src** directory? we currently don not have such a file...

src/helpers/utils.tsx
cKellyDesign: Excellent!

src/helpers/utils.tsx
cKellyDesign: Cool let's leave it as is then 👍

src/helpers/utils.tsx
```diff
@@ -60,6 +103,120 @@ export const sortFunction = (firstE1: SmsData, secondE1: SmsData): number => {
...
+  provinces: Location[],
+  districts: Location[],
+  communes: Location[],
+  villages: Location[]
+): FilterFunctionAndLocationLevel {
+  let locationFilterFunction: (smsData: SmsData) => boolean = () => {
+    return false;
+  };
+
+  let userLocationLevel = 4;
```
cKellyDesign: The location hierarchy is implied by the order of the function parameters, but it might be helpful to explicitly set the ordering in an array to make it iterable, like:
```
...
const locations = [provinces, districts, communes, villages]
let userLocationLevel = locations.length; // don't subtract 1 b/c location hierarchy from params isn't 0 based
...
```

src/helpers/utils.tsx
```diff
@@ -141,3 +306,133 @@ export function locationDataIsAvailable(
...
+
+/**
+ * Get a link to the HierrarchichalDataTable component
+ * @param {HIGH | LOW | NO} riskLevel string value that will be passed to hierarchichalDataTable
+ * as prop. representing that column should be coloured.
+ * @param {PREGNANCY | NBC_AND_PNC_CHILD | NBC_AND_PNC_WOMAN | NUTRITION | ''} module string representing the module
+ * @param {string} title title to be passed to hierarchichal data table as a prop
+ * @param {number} permissionLevel - number ranging from 0 - 4 that represents the permission level of the user.
+ * @param {string} locationId - the users location id.
+ */
```
cKellyDesign: @FrankApiyo should riskLevel and module have their own types? Just thinking it may be a bit less maintenance in the future as more levels / modules could be added.
```
export type RiskLevel = HIGH | LOW | NO;
export type ModuleType = PREGNANCY | NBC_AND_PNC_CHILD | NBC_AND_PNC_WOMAN | NUTRITION | '';
```

src/store/ducks/locations.ts
```diff
@@ -14,8 +14,30 @@ export interface Location {
...
+ * the OpenSRP authentication endpoint
+ */
+export interface UserLocation {
+  provider_name: string;
+  provider_contact: string;
+  provider_id: string;
+  location_id: string | any;
+  openmrs_identifier?: string;
+  location_name: string;
+}
```
cKellyDesign: Is it possible to get the location level from the OpenSRP endpoint? If not it might be worth requesting.

src/store/ducks/sms_events.ts
```diff
@@ -28,6 +29,10 @@ export interface SmsData {
   mother_symptoms: string;
 }
 
+export interface SmsDataByEventId {
+  [key: string]: SmsData;
+}
```
cKellyDesign: Doc-string please 🙏

src/store/ducks/sms_events.ts
```diff
@@ -155,7 +160,7 @@ export default function reducer(state: SmsState = initialState, action: SmsActio
  * @return { { [key: string] : SmsData}[] } - SmsData object[] as values, respective ids as keys
  */
 export function getSmsData(state: Partial<Store>): SmsData[] {
-  return values((state as any)[reducerName].smsData);
+  return Object.values((state as any)[reducerName].smsData);
```
cKellyDesign: Why the change? Feels like `values()` protects this function against `Object.values(undefined)`, ama?

src/store/ducks/sms_events.ts
FrankApiyo: This interface seems to have no use here... so I'm just going to remove it.

src/store/ducks/locations.ts
FrankApiyo: a request has been made already.

src/helpers/utils.tsx
```diff
@@ -60,6 +103,120 @@ export const sortFunction = (firstE1: SmsData, secondE1: SmsData): number => {
...
+}
+
+/**
+ *
+ * @param locationId location ID that we want to find in locations
+ * @param locations a list of Location objects from which we want to find a location ID
+ */
+export const locationIdIn = (locationId: string, locations: Location[]) => {
+  return locations.find((location: Location) => location.location_id === locationId);
+};
```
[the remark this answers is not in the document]
FrankApiyo: locations currently come from 4 different superset slices; each is lot larger than 15 values. so... not so many

src/helpers/utils.tsx
FrankApiyo: we could change ModuleType to `string`.

src/helpers/utils.tsx
```diff
@@ -28,11 +45,37 @@ export function oAuthUserInfoGetter(apiResponse: { [key: string]: any }): Sessio
...
+  const dataMap: FlexObject = {};
+  list.forEach((listElement: FlexObject) => {
+    if (!dataMap[listElement[field]]) {
+      dataMap[listElement[field]] = {
+        ...listElement,
+      };
+    }
+  });
+  return dataMap;
+}
```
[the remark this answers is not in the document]
FrankApiyo: yes. I think we should check for listElement...

src/helpers/utils.tsx
```diff
@@ -28,11 +45,37 @@ export function oAuthUserInfoGetter(apiResponse: { [key: string]: any }): Sessio
...
+
+/**
+ * Append a number suffix such as 'st' for 1 and 'nd' for 2 and so on.
+ * @param num
+ */
+export function getNumberSuffix(num: number): string {
```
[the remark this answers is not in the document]
FrankApiyo: @cKellyDesign  Since we are going to eventually have a monorepo here... I'm not sure where such reusable code should be moved to... js-tools or our packages here. 

cc: @moshthepitt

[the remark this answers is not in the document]
moshthepitt: @FrankApiyo when deciding whether to use OpenSRP packages or js-tools, we need to decide whether the reusable code is tightly coupled with OpenSRP or not. e.g. opensrp-server-service firmly belongs here while something like gatekeeper does not.

src/helpers/utils.tsx
```diff
@@ -60,6 +103,120 @@ export const sortFunction = (firstE1: SmsData, secondE1: SmsData): number => {
...
+export function getLocationId(userLocationData: UserLocation[], userUUID: string) {
+  const userDetailObj =
+    userLocationData.length &&
+    userLocationData.find(
+      (userLocationDataItem: UserLocation) => userLocationDataItem.provider_id === userUUID
+    );
+  if (userDetailObj) {
+    return userDetailObj.location_id;
+  }
+}
```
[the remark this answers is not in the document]
FrankApiyo: I tried to run like performance tests which were basically modified versions of this: https://stackoverflow.com/a/50844086/7682273 and find seems faster.

src/helpers/utils.tsx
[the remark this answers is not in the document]
cKellyDesign: Feels like adding suffixes to numbers is pretty generic.

FrankApiyo:
@moshthepitt since you are on leave and we needed to merge this, I dismissed that review.

Issue #113, opened 2019-11-21
Correct the description text shown on the Home page for `NBC & PNC module` and the `Nutrition` module.


FrankApiyo:
The text currectly shown on this modules is the same as that shown on the `pregnancy` module. This should be corrected to the same text as in the mock-ups.

PR #114, opened 2019-11-21
[DNM]Issue#113

FrankApiyo:
Fixes: #113

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "prop-types": "^15.7.2",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
+    "react-html-parser": "^2.0.2",
     "react-redux": "^7.1.0",
     "react-router": "^5.0.1",
     "react-router-dom": "^5.0.1",
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@types/node": "^12.6.8",
     "@types/react": "^16.8.23",
     "@types/react-dom": "^16.8.5",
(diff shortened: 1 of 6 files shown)


p-netm:
@FrankApiyo Please give the pr a better name and if possible write something about what this pr fixes.

FrankApiyo:
This is not ready for a review yet.

cKellyDesign:
@FrankApiyo is #113 the only issue being addressed in this PR?

FrankApiyo:
> @FrankApiyo is #113 the only issue being addressed in this PR?

yes @cKellyDesign. The reason why react-html-parser was installed is because some of the text we need to show on the compartment home may need parts of it to be bold and I thought this would be a pragmatic approach to take.

FrankApiyo:
> This is not ready for a review yet.

this is because this is something I need to ask @hmakandi

cKellyDesign:
@FrankApiyo yeah using `react-html-parser` is better than dangerously setting, I'm only hesitant because it hasn't been maintained in a few years (https://github.com/wrakky/react-html-parser/releases) and it's using react/-dom `^16.0.0`. 

> because some of the text we need to show on the compartment home may need parts of it to be bold

This is a common scenario so we do need to have a solution for this so for now let's keep `react-html-parser`, it may even be worth forking to see if updating react breaks anything :)

FrankApiyo:
@hmakandi @moshthepitt I'm going to close this as it's not required

PR #119, opened 2019-11-26
[DNM]Nutrition compartments cleanup

FrankApiyo:
This PR addresses the following issues in the nutrition-compartments branch:

- [x] Fix some tslint failures.

- [x] Add a new map function for the nutrition module.

- [x] show 'vietnam' in stead of no header breadcrumb when the user is at country level.

- [x]  Fixe a typo in the side menu by replacying `Logface` with `Log face`.


src/components/VillageData/index.tsx
```diff
@@ -200,4 +203,30 @@ export default class VillageData extends React.Component<Props, State> {
...
+    return (
+      <tr key={dataItem.event_id}>
+        <td className="default-width">
+          <Link to={`${getModuleLink(this.props.module)}/child_patient_detail/${dataItem.anc_id}`}>
+            {dataItem.anc_id}
+          </Link>
+        </td>
+        <td className="default-width">{getNumberOfDaysSinceDate(dataItem.EventDate)}</td>
+        <td className="default-width">{dataItem.health_worker_location_name}</td>
+        <td className="default-width">{dataItem.mother_symptoms}</td>
```
KipSigei: this should be `dataItem.child_symptoms`

FrankApiyo: yes. good catch!.

src/components/VillageData/index.tsx
```diff
@@ -200,4 +203,30 @@ export default class VillageData extends React.Component<Props, State> {
...
+        <td className="default-width">{dataItem.health_worker_location_name}</td>
```
KipSigei: also here, the context  is to know where the child was delivered according to the column header, so maybe we should use `dataItem.delivery_location`. Not sure though on the correct field name so we probably should get clarification from @hmakandi

FrankApiyo:
@KipSigei, comments made on https://github.com/OpenSRP/opensrp-web/pull/118 should really be addressed there instead of here, since these 2 PRs address different concerns they should not be mixed up. This prevents the danger of PRs getting overly large; when this happens they become difficult to review and some people in the team shy away from giving their opinions on them.

Issue #111, opened 2019-11-21
Tests to check the props passed to Highcharts.


FrankApiyo:
The current test for the `WeightAndHeightChart` component mock Hicharts because of an error thrown if an attempt is made to render `Higcharts` using *jest*. No checks are made for the props passed to the actual chart component. For the tests on the `WeightAndHeightChart` to be effective, this checks should  be made.

PR #121, opened 2019-12-02
Issue#111 check that hicharts receives correct props

FrankApiyo:
Fixes: #111


src/components/VillageData/tests/index.test.tsx
```diff
@@ -78,8 +76,7 @@ describe('components/VillageData/pregnancyMapFunction', () => {
       </Provider>
     );
     const instance = wrapper.find('VillageData').instance() as VillageData;
-    expect(instance.pregnancyMapFunction(villageDataProps.smsData[2])).toMatchSnapshot(
-      'pregnancyMapFunctionOutput'
-    );
+    const tableRows = shallow(instance.pregnancyMapFunction(villageDataProps.smsData[2]));
+    expect(tableRows.find('td[className="default-width"]').length).toEqual(7);
```
FrankApiyo: These tests have been changed because one of the `<td>` elements returned by each of these functions tested here has a value that is generated based on today's date, so even if the snapshots were updated now... tomorrow they would fail again.

src/components/WeightAndHeightChart/test/index.test.tsx
```diff
@@ -51,5 +60,23 @@ describe('WeightAndHeightChart', () => {
...
+          <WeightAndHeightChart weights={weights} {...defaultProps} />
+        </Router>
+      </Provider>
+    );
+    jest.runAllTimers();
+    const argument1 = (Highcharts.chart as jest.MockedFunction<any>).mock.calls[0][0];
+    const argument2 = (Highcharts.chart as jest.MockedFunction<any>).mock.calls[0][1];
+    expect(argument1).toEqual('wrapper-id');
+    delete argument2.tooltip.formatter;
+    expect(argument2).toEqual(chartArgument);
```
cKellyDesign: Instead of deleting the `formatter`, can we override it as a mocked function on the fixture after we've imported it?

FrankApiyo: @cKellyDesign this function is really hard to mock coz of the fact that it's not exported. and it's hard to export because of how it's defined.

FrankApiyo: but' I've resorted to overriding `formatter` in `chartArgument.tooltip` with what was actually passed to `Highcharts.chart` and ensuring that it's actually a function because.

PR #122, opened 2019-12-02
Change to new infrastructure

### LINKED ISSUE #7: To Monorepo or not?
I'm thinking that it may be a good idea to consider restructuring this repo as follows (yes it is a monorepo):

```
opensrp-web/
|-- server/ (probably an express.js tiny backend)
|-- client/ (the CRA frontend code)
|-- packages/ (all the different OpenSRP modules)
```

Let me go through the parts above and their purpose:

- **server**: this would be a tiny tiny express.js backend that may be useful for a lot of things as previously discussed in https://github.com/onaio/reveal-frontend/issues/39 and https://github.com/onaio/reveal-frontend/issues/156.  The server will only be used when it is actually needed
- **client**: a really stripped down CRA ([create react app](https://facebook.github.io/create-react-app/)) that has the bare minimum structure needed for a client side react app (maybe just index.js and app.js) and loads the rest from `packages`.  It should be possible to deploy this independently (i.e. without the `server` above)
- **packages**: herein will be code for the various OpenSRP modules (structured similarly to [js-tools](https://github.com/onaio/js-tools)).  The idea is to make it possible for an OpenSRP web deployment to use only the modules that it needs in as modular and DRY a way as possible

### pros

- It would be easier to deploy OpenSRP web apps that have only the compiled javascript that they need and no more (this happens to be my main motivation in proposing this structure)
- It would be easier to set up the express.js mini server as the backend of the react app(s)
- Having everything in one repo makes the life of devs easier, arguably

### cons

- All the usual cons of having a monorepo

moshthepitt:
This fixes #7 

- [x] Switch to a monorepo managed by Lerna
- [x] Move current contents of master branch to `/clients/core`
- [x] Create packages directory
- [x] Add OpenSRP server service package
- [x] Replace tslint with eslint
- [x] Make sure automated tests work


moshthepitt:
TODO:

- [ ] Fix linting errors clients/core
- [ ] Review and fix any issues with @opensrp/server-service
- [ ] Ensure automated tests are run for all clients and packages

FrankApiyo:
> @FrankApiyo I am not usually for the idea of turning rules off. Why is this necessary?

I've turned off two rules: 

1. @typescript-eslint/explicit-function-return-type:
This rule requires that all functions have return types explicitly specified. I'm not sure if this is necessary for all the component life cycle methods and the arrow functions.
2. @typescript-eslint/no-inferrable-types:
This rule requires that types that can be inferred from initializing a variable are not specified explicitly, I don't think we should have a rule that reduces the degree to which code is self documenting.

We could also instead, for starters set these to `warning` instead of `off` so that they warn the devs when they break these rules but are not marked as linting errors.  

CC: @moshthepitt @p-netm @cKellyDesign

moshthepitt:
@FrankApiyo please set them to warning

FrankApiyo:
should we remove `.vscode` from `.gitignore`? 

I think we should our repo should be able to do things like recommend extensions automatically and also configure es-lint automatically.

cc: @p-netm @moshthepitt @redoy11 @cKellyDesign

moshthepitt:
> should we remove .vscode from .gitignore?

I think we should keep it if it doesn't contain anything specific to one's local machine

.circleci/config.yml
```diff
@@ -50,13 +50,16 @@ jobs:
...
       # run tests!
-      - run: 
-          name: Run test with coverage
-          command: yarn test --verbose --coverage=true --forceExit --detectOpenHandles
+      - run:
+          name: Run test with coverage in clients/core
+          command: cd clients/core && yarn test --verbose --coverage=true --forceExit --detectOpenHandles && cd ../../
+      - run:
+          name: Run test with coverage in packages
+          command: cd packages/ && yarn test --verbose --coverage=true --forceExit --detectOpenHandles && cd ../
```
moshthepitt: Do you need to cd into packages to run this?  I think not, but please confirm.

FrankApiyo: Yes. we do not need to cd into the packages to run the test. this will be reverted.

FrankApiyo:
Todo: add  documentation on:

- [ ] running tests for a single package/client
- [ ] running tests for the whole repo
- [ ] creating a new package
- [ ] creating a new client

clients/core/src/store/index.ts
```diff
@@ -2,10 +2,11 @@ import { connectReducer, getConnectedStore } from '@onaio/connected-reducer-regi
 import { gateKeeperReducer, gateKeeperReducerName } from '@onaio/gatekeeper';
 import reducerRegistry, { combine, Registry } from '@onaio/redux-reducer-registry';
 import session, { reducerName as sessionReducer } from '@onaio/session-reducer';
+import { Reducer } from 'redux';
 
 /** Initial reducers in the reducer registry */
 const defaultReducers: Registry = {
-  router: connectReducer as any /** Dirty hack because connectRouter LocationChangeAction does not extend Redux.AnyAction */,
+  router: connectReducer as Reducer,
```
cKellyDesign: Nice!

FrankApiyo:
This roadmap :point_right:  https://github.com/typescript-eslint/typescript-eslint/blob/324f155fc2c4618e5b844e9a481607a08ed821da/packages/eslint-plugin/ROADMAP.md says that; `object-literal-sort-keys`, a rule that we have used a lot is currently not available for ESlint.

.circleci/config.yml
```diff
@@ -50,13 +29,16 @@ jobs:
...
-          name: lint Ts
-          command: yarn lint-ts
-      # run tests!
-      - run: 
+      - run:
+          name: Run linter
+          command: yarn lint
+      # run tests for client/score!
+      - run:
+          name: Run test with coverage
```
p-netm: `Run test with coverage for clients` maybe ?

clients/core/package.json
```diff
@@ -0,0 +1,103 @@
...
+    "redux": "^4.0.4",
+    "redux-thunk": "^2.3.0",
+    "seamless-immutable": "^7.1.4",
+    "snapshot-diff": "^0.5.2",
+    "typescript": "^3.5.3"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "cp ./.env ../../.env && cd ../../ && yarn test clients/core --verbose --collectCoverage=true && cd clients/core",
```
p-netm: Am apprehensive about the first 2 shell commands, seems like it implies we will be having a single env file across the client applications. 

While am here, il'd like to ask if its a better idea to localize the testing setup to each of the clients packages?

package.json
```diff
@@ -1,134 +1,97 @@
...
-      "!**/node_modules/**",
-      "!**/coverage/**",
-      "!**/serviceWorker.ts",
-      "!**/setupTests.ts",
-      "!src/index.tsx",
-      "!**/*.d.ts"
+    "name": "root",
+    "private": true,
+    "scripts": {
+        "test": "jest",
```
p-netm: Can we add another test command along the lines of `"test:watch": "jest --watch"`

README.md
```diff
@@ -65,3 +74,126 @@ See [contribution guidelines](https://github.com/OpenSRP/opensrp-web/blob/docs/d
 ## Code Quality
 
 See [guidelines](docs/codeQuality.md) on recommended coding conventions for this project.
+
+## Creating a New Client
+
+```sh
+cd clients
+npx create-react-app my-new-client --typescript
```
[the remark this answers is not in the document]
FrankApiyo: I think so too :smile:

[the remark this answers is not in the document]
FrankApiyo: Fixing that.

README.md
```diff
@@ -65,3 +74,126 @@ See [contribution guidelines](https://github.com/OpenSRP/opensrp-web/blob/docs/d
...
+mkdir my-new-package && cd my-new-package
+```
+
+Then we create a new package.json by running yarn init:
+
+```sh
+yarn init
+```
+
+The new name should follow our NPM Org scope e.g. `@onaio`
```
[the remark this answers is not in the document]
FrankApiyo: Yeah... I was wondering about this too.

README.md
```diff
@@ -65,3 +74,126 @@ See [contribution guidelines](https://github.com/OpenSRP/opensrp-web/blob/docs/d
...
+
+Add a `.prettierrc` file such as this one.
+
+```sh
+{
+    "printWidth": 100,
+    "singleQuote": true,
+    "trailingComma": "es5"
+}
+```
```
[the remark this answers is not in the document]
FrankApiyo: Makes sense...

clients/core/package.json
FrankApiyo: we could do that... I see that as a plausible alternative we have to localize the testing setup to each of the client packages as you put it.

Could we consider this a future addition and maybe have a discussion on it.

PR #124, opened 2019-12-03
Addition of Household dux module

prodhan29:
Ducks module of clients and household will be added along with its test code.

### DIFF
--- a/src/store/ducks/clients.ts
+++ b/src/store/ducks/clients.ts
@@ -127,14 +127,6 @@ export function getClientsById(state: Partial<Store>): { [key: string]: Client }
   return (state as any)[reducerName].clientsById;
 }
 
-/** get clients ids in an array
- * @param {Partial<Store>} state - the redux store
- * @return {string[]} - clients ids as an array of strings
- */
-export function getClientsIdArray(state: Partial<Store>): string[] {
-  return keys(getClientsById(state));
-}
-
 /** gets clients as an array of clients objects
  * @param {Partial<Store>} state - the redux store
  * @return {Client[]} - an array of clients objs
--- a/src/store/ducks/clients.ts
+++ b/src/store/ducks/clients.ts
@@ -31,19 +31,14 @@ export interface Client {
   _rev: string;
 }
 
-/** Interface for household object same as client */
-export type Household = Client;
-
 // actions
 
 /** CLIENTS_FETCHED action type */
 export const CLIENTS_FETCHED = 'opensrp/reducer/clients/CLIENTS_FETCHED';
 /** REMOVE_CLIENTS action type */
 export const REMOVE_CLIENTS = 'opensrp/reducer/clients/REMOVE_CLIENTS';
-/** HOUSEHOLDS_FETCHED action type */
-export const HOUSEHOLDS_FETCHED = 'opensrp/reducer/clients/HOUSEHOLDS_FETCHED';
-/** REMOVE_HOUSEHOLDS action type */
-export const REMOVE_HOUSEHOLDS = 'opensrp/reducer/clients/REMOVE_HOUSEHOLDS';
+/** SET NAVIGATION PAGE */
+export const SET_NAVIGATION_PAGE = 'opensrp/reducer/clients/SET_NAVIGATION_PAGE';
 
 /** interface for authorize action */
 export interface FetchClientsAction extends AnyAction {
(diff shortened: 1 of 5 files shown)


src/store/ducks/households.ts
```diff
@@ -0,0 +1,141 @@
...
+  return (state as any)[reducerName].householdsById;
+}
+
+/** gets households as an array of households objects
+ * @param {Partial<Store>} state - the redux store
+ * @return {Household[]} - an array of households objs
+ */
+export function getHouseholdsArray(state: Partial<Store>): Household[] {
+  return values(getHouseholdsById(state));
+}
```
p-netm: nitpick: will we be needing this?

src/store/ducks/clients.ts
```diff
@@ -31,19 +31,14 @@ export interface Client {
...
 /** CLIENTS_FETCHED action type */
 export const CLIENTS_FETCHED = 'opensrp/reducer/clients/CLIENTS_FETCHED';
 /** REMOVE_CLIENTS action type */
 export const REMOVE_CLIENTS = 'opensrp/reducer/clients/REMOVE_CLIENTS';
-/** HOUSEHOLDS_FETCHED action type */
-export const HOUSEHOLDS_FETCHED = 'opensrp/reducer/clients/HOUSEHOLDS_FETCHED';
-/** REMOVE_HOUSEHOLDS action type */
-export const REMOVE_HOUSEHOLDS = 'opensrp/reducer/clients/REMOVE_HOUSEHOLDS';
+/** SET NAVIGATION PAGE */
+export const SET_NAVIGATION_PAGE = 'opensrp/reducer/clients/SET_NAVIGATION_PAGE';
```
p-netm: what implementation detail would an action with this type handle?

src/store/ducks/households.ts
```diff
@@ -0,0 +1,141 @@
...
+
+/** Interface for household object same as client */
+export type Household = Client;
+
+/** HOUSEHOLDS_FETCHED action type */
+export const HOUSEHOLDS_FETCHED = 'opensrp/reducer/households/HOUSEHOLDS_FETCHED';
+/** REMOVE_HOUSEHOLDS action type */
+export const REMOVE_HOUSEHOLDS = 'opensrp/reducer/households/REMOVE_HOUSEHOLDS';
+/** SET_TOTAL_RECORDS action type */
+export const SET_TOTAL_RECORDS = 'opensrp/reducer/households/SET_TOTAL_RECORDS';
```
p-netm: do we also need this in the clients dux module.

redoy11: Well. This is used to set the actual total records received from the server.

src/store/ducks/clients.ts
redoy11: Well. This would have been used to set the current page of the table list in the redux store, but since we agreed on storing the current page number in the component state, we can safely remove this.

prodhan29: it was previously used to hold the pageNumber in clients page. we had a discussion on it and decided to manage the pageNumber in the component state. 
it is no longer needed here. so i am removing it.

src/store/ducks/households.ts
prodhan29: yes, we need this to get households in an array. so that we can easily loop over it display households in householdList component. though householdList component is not in this branch.

src/store/ducks/households.ts
p-netm: @prodhan29 yeah i get that, what i am asking, is that this is not in the clients dux module, is it supposed to also be in the ducks module?

src/store/ducks/clients.ts
p-netm: ok, cool.

src/store/ducks/households.ts
p-netm: @prodhan29  sorry, i meant to flag this `getClientsIdArray` function [here](https://github.com/OpenSRP/opensrp-web/blob/8cc0aba23cb3e264c058d8ecd15a036aa42f5726/src/store/ducks/clients.ts#L159)

src/store/ducks/households.ts
redoy11: @p-netm  Sorry about earlier. Yes, this is also needed for the clients dux module.

src/store/ducks/households.ts
prodhan29: function [getClientsIdArray](https://github.com/OpenSRP/opensrp-web/blob/8cc0aba23cb3e264c058d8ecd15a036aa42f5726/src/store/ducks/clients.ts#L159)
this method is used only in the test files. we may need it later

p-netm: we should try and minimize unused code as possible.

prodhan29: i removed the code for now.

src/store/ducks/households.ts
p-netm: @prodhan29 :point_up:

prodhan29: @p-netm i think we should create an another pr to add `total_records` for clients dux module. if i  add that in this pr. it will get long like the previous one.

Issue #137, opened 2019-12-18
The back button on patientDetail and childDetail pages does not work on firefox


FrankApiyo:
![Screenshot from 2019-12-18 15-10-32 (1)](https://user-images.githubusercontent.com/25260439/71086150-2a890080-21aa-11ea-82a6-c757b0517a3e.png)

It is possible that this does not work on many other browsers too.

FrankApiyo:
@KipSigei, @cKellyDesign I'm considering using this :point_right:  https://github.com/hinok/react-router-last-location to solve this issue. Thoughts?

cKellyDesign:
Have we tried solving this with React's `history`? Something like:
```
import { createBrowserHistory } from 'history';
const history = createBrowserHistory();
history.goBack();
```
Is this a different history module than is passed in through the router props?

☝️@FrankApiyo If this doesn't work then go for it with `react-router-last-location`

FrankApiyo:
> Have we tried solving this with React's `history`? Something like:
> 
> ```
> import { createBrowserHistory } from 'history';
> const history = createBrowserHistory();
> history.goBack();
> ```
> 
> Is this a different history module than is passed in through the router props?
> 
> point_up@FrankApiyo If this doesn't work then go for it with `react-router-last-location`

@cKellyDesign, this approach did not work too... so I'm going to go with `react-router-last-location`

PR #142, opened 2019-12-23
Ensure back button on patient detail page works on all browsers.

FrankApiyo:
Fixes: #137  and also refactors the `PatientDetail` compartment to become a functional component with state.

The following changes have been made in this PR:

- Install `react-router-last-location`
- Rename report table `isNutrition` prop to `isChild`
- Remove `ChildPatientDetail` component and replace with a more general `PatientDetail` component.
- Refactor the `PatientDetails` component to use hooks


package.json
```diff
@@ -46,6 +46,7 @@
     "react-redux": "^7.1.0",
     "react-router": "^5.0.1",
     "react-router-dom": "^5.0.1",
+    "react-router-last-location": "^2.0.1",
```
cKellyDesign: 👍

src/components/ReportTable/index.tsx
```diff
@@ -130,7 +131,7 @@ export const getEventsPregnancyArray = (
 class ReportTable extends Component<Props, State> {
   public static getDerivedStateFromProps(props: Props, state: State) {
     return {
-      pregnancyEventsArray: getEventsPregnancyArray(props.singlePatientEvents, props.isNutrition),
+      pregnancyEventsArray: getEventsPregnancyArray(props.singlePatientEvents, props.isChild),
```
cKellyDesign: Nice generalization!

src/containers/ChildPatientDetails/index.tsx
```diff
@@ -1,263 +0,0 @@
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
cKellyDesign: Is this container no longer used? Was it replaced or generalized or something?

src/containers/Compartments/index.tsx
```diff
@@ -130,11 +130,13 @@ export const Compartments = ({
   useEffect(() => {
     removeFilterArgs();
     fetchData();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
cKellyDesign: Can we disable this lint rule in the lint config? Seems like this would be needed every time we use the `useEffect()` pattern which we are doing more and more.

src/containers/PatientDetails/index.tsx
```diff
@@ -15,183 +16,170 @@ import {
...
+  useEffect(() => {
+    setFilteredData(
+      sortByEventDate(
+        filterByPatientId({
+          patientId,
+          smsData,
+        })
+      )
+    );
+  }, [patientId, smsData]);
```
cKellyDesign: Why doesn't this `useEffect` need the lint ignores?

src/containers/PatientDetails/index.tsx
```diff
@@ -15,183 +16,170 @@ import {
...
 
-interface State {
-  filteredData: SmsData[];
-}
-
-const defaultProps: Partial<Props> = {
-  isNutrition: false,
-  patientId: 'none',
-  smsData: [],
+export const PatientDetails = ({ isChild = false, patientId = 'none', smsData = [] }: Props) => {
```
cKellyDesign: 👏 Excellent, love to see more functional components :)

src/containers/PatientDetails/index.tsx
```diff
@@ -15,183 +16,170 @@ import {
...
+    if (reversedFilteredData[data]) {
+      for (const field in statusFields) {
+        if ((reversedFilteredData[data] as any)[statusFields[field]]) {
+          return (reversedFilteredData[data] as any)[statusFields[field]];
         }
       }
     }
-    return 'no risk category';
   }
+  return 'no risk category';
```
cKellyDesign: Let's move messages like this into `constants` since they are display strings and my need to be translatable in the future.

src/containers/PatientDetails/index.tsx
```diff
@@ -15,183 +16,170 @@ import {
...
+function getCurrentLocation(filteredData: SmsData[]): string {
+  const reversedFilteredData: SmsData[] = [...filteredData];
+  reversedFilteredData.reverse();
+  for (const data in reversedFilteredData) {
+    if (reversedFilteredData[data].health_worker_location_name) {
+      return reversedFilteredData[data].health_worker_location_name;
     }
-    return 'could not find any location';
   }
+  return 'could not find any location';
```
cKellyDesign: Same request as above.

src/containers/PatientDetails/index.tsx
```diff
@@ -15,183 +16,170 @@ import {
...
-          { label: RESIDENCE, value: this.getCurrentLocation() },
-        ] as LabelValuePair[]);
-    return basicInformationProps;
+function getPreviousPregnancyRisk(filteredData: SmsData[]): string {
+  const reversedFilteredData: SmsData[] = [...filteredData];
+  reversedFilteredData.reverse();
+  if (reversedFilteredData[1]) {
+    return reversedFilteredData[1].logface_risk;
+  } else {
+    return 'no risk';
```
cKellyDesign: Same request as above, these aren't actually magical strings but they are display strings

src/containers/PatientDetails/index.tsx
```diff
@@ -15,183 +16,170 @@ import {
...
-const mapStateToprops = (state: any, ownProps: any) => {
-  const patient_id = ownProps.match.params.patient_id;
-  const result = {
-    isNutrition: ownProps.match.url.includes(NUTRITION.toLowerCase()),
-    patientId: patient_id,
+const mapStateToprops = (
+  state: any,
+  ownProps: RouteComponentProps<{
+    patient_id: string;
+  }> & { isChild: boolean }
```
cKellyDesign: Would it be useful to define an interface for this?

src/helpers/utils.tsx
```diff
@@ -285,26 +285,40 @@ export const getProvince = (
...
+ * @param {PatientIDAndSmsData} patientIdAndSmsData an object with the patient id and smsData
+ * @return {SmsData[]} filtered smsData
+ */
+export const filterByPatientId = (patientIdAndSmsData: PatientIDAndSmsData): SmsData[] => {
+  return [...patientIdAndSmsData.smsData].filter((dataItem: SmsData): boolean => {
+    return dataItem.anc_id
+      .toLocaleLowerCase()
+      .includes(patientIdAndSmsData.patientId.toLocaleLowerCase());
+  });
+};
```
cKellyDesign: Is this function still sorting? If not then let's update the doc-string pls.

src/containers/ChildPatientDetails/index.tsx
FrankApiyo: yeah, the `patientDetail` container is general enough now to replace it.

src/helpers/utils.tsx
FrankApiyo: not sorting anymore; updated

src/containers/PatientDetails/index.tsx
FrankApiyo: we can assign this type to a variable like :
```
type RouterProps = RouteComponentProps<{
  patient_id: string;
}> & { isChild: boolean };
```

src/containers/Compartments/index.tsx
FrankApiyo: I'll create an issue for this because if we disabled this rule in the lint configs we would have to change the `.eslintrc` file. This file has been removed in the `master` branch because it has been replaced by `.eslintrc.js`, I'd preffer disabling it there.

FrankApiyo: I should add that the reason why I keep disabling this rule here is because adding some variables in the dependency arrays of these hooks causes the logic to change in very nuanced and confusing ways.

src/containers/PatientDetails/index.tsx
FrankApiyo: @cKellyDesign  that's because all the variables used within this hook have been included in it's dependency array. I guess that's why the rule is called `react-hooks/exhaustive-deps`; because it ensures that the dependency array is exhaustive in the sense that all variables used within the hook are included in the dependency array. :smiley:

src/containers/Compartments/index.tsx
FrankApiyo: @cKellyDesign on second thought, I don't think we should disable this rule, since:
-  it only gives a warning
- it comes highly recommended by the react team

moshthepitt:
It would be cool to change the title of this PR to something more descriptive.

FrankApiyo:
I'm going to close this PR and replace it with one that merges to the `miecd_vietnam` branch as branch `Issue#130` was already merged to `miecd_vietnam`.

FrankApiyo:
> It would be cool to change the title of this PR to something more descriptive.

@moshthepitt I've now added more description.

src/containers/PatientDetails/index.tsx
cKellyDesign: Awesome thanks makes sense!

src/containers/Compartments/index.tsx
cKellyDesign: Good call :)

Issue #144, opened 2019-12-23
Nutrition compartments children under 5 data circle card working incorrectly


FrankApiyo:
Nutrition compartments children under 5 data circle card should not show any data on the hierarchical data table when clicked.

hmakandi:
Closing this though need to clarify that the item should be should show data on the hierarchical table. Currently works.

PR #145, opened 2019-12-23
Add filter args to the nutrition DataCircleCards

FrankApiyo:
Fixes: #144

### DIFF
--- a/src/containers/Compartments/index.tsx
+++ b/src/containers/Compartments/index.tsx
@@ -450,7 +450,7 @@ export const Compartments = ({
 };
 
 /**
- * filter function for smsData based on data_of_birth field
+ * filter function for smsData based on date_of_birth field
  * @param {SmsData} dataItem - SmsData item
  * @returns {boolean}
  */
@@ -459,7 +459,7 @@ const childrenUnder2FilterFunction = (dataItem: SmsData): boolean => {
 };
 
 /**
- * filter function for smsData based on data_of_birth field
+ * filter function for smsData based on date_of_birth field
  * @param {SmsData} dataItem - SmsData item
  * @returns {boolean}
  */
--- a/src/containers/Compartments/index.tsx
+++ b/src/containers/Compartments/index.tsx
@@ -489,7 +489,7 @@ const getNumberOfSmsWithRisk = (risk: string, smsData: SmsData[], field: string)
 /**
  * @param {number} n number of weeks future
  */
-const filterByDateInNextNWeeks = (n: number) => {
+export const filterByDateInNextNWeeks = (n: number) => {
   return (dataItem: SmsData) => {
     return (
       Date.parse(dataItem.lmp_edd) - Date.now() > 0 &&
--- a/src/containers/Compartments/index.tsx
+++ b/src/containers/Compartments/index.tsx
@@ -26,6 +26,7 @@ import {
 } from '../../constants';
 import {
   buildHeaderBreadCrumb,
+  convertMilisecondsToYear,
   fetchData,
   getFilterFunctionAndLocationLevel,
   getLocationId,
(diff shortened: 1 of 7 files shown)


FrankApiyo:
> Looks good to me, are there tests for this scenario?

@cKellyDesign, adding tests

moshthepitt:
@FrankApiyo so this component uses a lot of functions (the various filter functions and others).  In my opinion we should prefer to make things less tightly decoupled.  One way to do this is by having all these various functions fed into this component as props (with default values) - because I think loose coupling is a good principle for such components and because I think we should try to make this component more [open for extension but closed for modification](https://stackify.com/solid-design-open-closed-principle/).   Curious if you think that this is a good idea.

FrankApiyo:
> I know they are supposed to be REALLY constant constants, but I feel very uncomfortable with the magic numbers used here (0, 2, and 5 - possibly others). Do you think that it is a good idea to get these from configurable variables?

@moshthepitt, I think instead of writing these two functions that filter `SmsData` objects in preset ranges specified by this constants(0, 2 and 5), A single function that allows you to decide what range you would like to filter by would server the purpose of these two better.

FrankApiyo:
> @FrankApiyo so this component uses a lot of functions (the various filter functions and others). In my opinion we should prefer to make things less tightly decoupled. One way to do this is by having all these various functions fed into this component as props (with default values) - because I think loose coupling is a good principle for such components and because I think we should try to make this component more [open for extension but closed for modification](https://stackify.com/solid-design-open-closed-principle/). Curious if you think that this is a good idea.

@moshthepitt, loose coupling these functions with the component is a good idea. Although the LOE for refactoring this component to take these functions are props right now may be a little too high. Although, I don think most of these functions can be made more generic by removing the magic numbers used in them.

src/containers/Compartments/test/__snapshots__/index.test.tsx.snap
```diff
@@ -335,15 +335,731 @@ Array [
 ]
 `;
 
-exports[`Compartments must render correctly: compartment title 1`] = `
+exports[`Compartments must render correctly for NBC_AND_PNC module 1`] = `
```
moshthepitt: Let us make this snapshot smaller please

src/containers/Compartments/index.tsx
```diff
@@ -320,21 +323,14 @@ export const Compartments = ({
   ] = useState<null | NutritionDataCircleCardProps>(null);
 
   useEffect(() => {
-    const childrenUnder2 = filteredData.filter((dataItem: SmsData) => {
-      return new Date().getFullYear() - new Date(dataItem.date_of_birth).getFullYear() < 2;
-    });
+    const childrenUnder2 = filteredData.filter(childrenAgeRangeFilterFunction(0, 2));
```
moshthepitt: This function call is repeated.  What do you think of assigning it to a variable to avoid that?

src/containers/Compartments/index.tsx
```diff
@@ -320,21 +323,14 @@ export const Compartments = ({
...
 
-    const childrenUnder5 = filteredData.filter((dataItem: SmsData) => {
-      return (
-        new Date().getFullYear() - new Date(dataItem.date_of_birth).getFullYear() < 5 &&
-        new Date().getFullYear() - new Date(dataItem.date_of_birth).getFullYear() > 2
-      );
-    });
+    const childrenUnder5 = filteredData.filter(childrenAgeRangeFilterFunction(2, 5));
```
moshthepitt: This function call is repeated.  What do you think of assigning it to a variable to avoid that?

FrankApiyo: makes sense; done!

PR #146, opened 2019-12-31
Add Component household-list

prodhan29:
- Add Household-list component
- Test case for household-list

### DIFF
--- a/clients/core/src/containers/Households/List/index.tsx
+++ b/clients/core/src/containers/Households/List/index.tsx
@@ -1,0 +1,191 @@
+import reducerRegistry from '@onaio/redux-reducer-registry';
+import { some } from 'lodash';
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { Link } from 'react-router-dom';
+import { Col, Row, Table } from 'reactstrap';
+import { Store } from 'redux';
+import Loading from '../../../components/page/Loading';
+import SearchBox from '../../../components/page/SearchBox/SearchBox';
+import { OPENSRP_HOUSEHOLD_ENDPOINT, PAGINATION_SIZE } from '../../../configs/env';
+import {
+  householdClientType as clientType,
+  householdSearchPlaceholder as searchPlaceholder,
+} from '../../../constants';
+import { FlexObject } from '../../../helpers/utils';
+import { OpenSRPService } from '../../../services/opensrp';
+import householdsReducer, {
+  fetchHouseholds,
+  getHouseholdsArray,
+  getTotalRecords,
+  Household,
+  reducerName as householdsReducerName,
+  removeHouseholds,
+  setTotalRecords,
+} from '../../../store/ducks/households';
+import './householdList.css';
+
+/** register the households reducer */
+reducerRegistry.register(householdsReducerName, householdsReducer);
+
+/** props Interface for the householdList component */
+export interface HouseholdListProps {
+  opensrpService: typeof OpenSRPService;
+  householdsArray: Household[];
+  totalRecordsCount: number;
+  fetchHouseholdsActionCreator: typeof fetchHouseholds;
+  setTotalRecordsActionCreator: typeof setTotalRecords;
+  removeHouseholdsActionCreator: typeof removeHouseholds;
+}
+
+export interface HouseholdListState {
+  loading: boolean;
+  pageNumber: string;
+}
+
+export const defaultHouseholdListState: HouseholdListState = {
+  loading: true,
+  pageNumber: '1',
+};
+
+/** default props for the householdList component */
+export const defaultHouseholdListProps: HouseholdListProps = {
+  fetchHouseholdsActionCreator: fetchHouseholds,
+  householdsArray: [],
+  opensrpService: OpenSRPService,
+  removeHouseholdsActionCreator: removeHouseholds,
+  setTotalRecordsActionCreator: setTotalRecords,
+  totalRecordsCount: 0,
+};
+
+/** Display the Household list  */
+class HouseholdList extends React.Component<HouseholdListProps, HouseholdListState> {
+  public static defaultProps: HouseholdListProps = defaultHouseholdListProps;
+
+  public constructor(props: HouseholdListProps) {
+    super(props);
+    this.state = defaultHouseholdListState;
+  }
+  public async componentDidMount() {
+    this.getHouseholdList();
+  }
+
+  public render() {
+    /** render loader if there are no households in state */
+    const { householdsArray, totalRecordsCount } = this.props;
+    if (this.state.loading) {
+      return <Loading />;
+    } else {
+      return (
+        <div>
+          <h3 className="household-title"> Household ({totalRecordsCount}) </h3>
+          <Row>
+            <Col md={5}>
+              <div className="household-search-bar">
+                <SearchBox searchCallBack={this.search} placeholder={searchPlaceholder} />
+              </div>
+            </Col>
+          </Row>
+          <Row>
+            <Col>
+              <Table className="shadow-sm">
+                <thead>
+                  <tr>
+                    <th>HH ID Number</th>
+                    <th>Family Name</th>
+                    <th>Head of Household</th>
+                    <th>Phone</th>
+                    <th>Registered Date</th>
+                    <th>Members</th>
+                    <th>Actions</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  {householdsArray.map((household: Household) => {
+                    return (
+                      <tr key={household.baseEntityId}>
+                        <td>{household.baseEntityId}</td>
+                        <td>{household.lastName}</td>
+                        <td>{household.attributes.HouseholdHead}</td>
+                        <td>{household.attributes.phoneNumber}</td>
+                        <td>{household.dateCreated}</td>
+                        <td className="members-table-field">{household.attributes.memberCount}</td>
+                        <td>
+                          <Link to={`/household/profile/${household.baseEntityId}/`}>View</Link>
+                        </td>
+                      </tr>
+                    );
+                  })}
+                </tbody>
+              </Table>
+            </Col>
+          </Row>
+        </div>
+      );
+    }
+  }
+
+  private search = (searchText: string) => {
+    this.getHouseholdList({ searchText });
+  };
+
+  private async getHouseholdList(extraParams: FlexObject = {}) {
+    const params = {
+      clientType,
+      pageNumber: this.state.pageNumber,
+      pageSize: PAGINATION_SIZE,
+      ...extraParams,
+    };
+    const {
+      fetchHouseholdsActionCreator,
+      setTotalRecordsActionCreator,
+      removeHouseholdsActionCreator,
+      opensrpService,
+    } = this.props;
+    const hosueholdService = new opensrpService(`${OPENSRP_HOUSEHOLD_ENDPOINT}`);
+    const response = await hosueholdService.list(params);
+    removeHouseholdsActionCreator();
+    this.setState(
+      {
+        ...this.state,
+        loading: false,
+      },
+      () => {
+        fetchHouseholdsActionCreator(response.clients);
+        setTotalRecordsActionCreator(response.total);
+      }
+    );
+  }
+}
+
+export { HouseholdList };
+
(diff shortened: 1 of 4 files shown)


FrankApiyo:
@prodhan29  Is there an issue for this?

prodhan29:
> @prodhan29 Is there an issue for this?

this [PR](https://github.com/OpenSRP/opensrp-web/pull/84) was divided into small 5 different pr. 

**Add Component household-list**  is one of it. i think there is no individual issue for creating household-list-component.

clients/core/src/containers/Households/List/tests/index.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+        };
+        const wrapper = mount(
+            <Provider store={store}>
+                <HouseholdList {...props} />
+            </Provider>,
+        );
+        expect(toJson(wrapper)).toMatchSnapshot();
+        wrapper.unmount();
+    });
+});
```
p-netm: I think we can do better at coverage here by:
1. testing it renders correctly: basically check the rendered output if it has crucial text or html nodes that prove the page rendered correctly,
2. for regression purposes we also want to leave a test that makes sure the component is making the correct calls.
3. For the same purposes we also want to leave a test that spies on the selectors and makes sure that they are called with the right arguments.
4. I also think a simulation like so would be in-order: try simulating getting data for at-least 2 `pageNumbers` in the service calls and test that the response in both cases is consumed by the container as intended.

The 4th one seems to be an interesting(really challenging) one , but it think its necessary, since at the moment the result got from an api call the the first page is different to responses from subsequent api calls to other pages

Also, please pay attention to the auto-generated snap files, scrutinize them carefully. Among other things, they can come in handy in pointing out logic holes in your test, as well as helping in debugging

clients/core/src/containers/Households/List/tests/index.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+            totalRecordsCount: 0,
+        };
+        const wrapper = mount(
+            <Provider store={store}>
+                <HouseholdList {...props} />
+            </Provider>,
+        );
+        expect(toJson(wrapper)).toMatchSnapshot();
+        wrapper.unmount();
+    });
```
p-netm: I don't think this test really exercises whether the component works with redux store,...One thing you want to make sure is that you are rendering the correct component(the container); dispatch some state to the store and then inspect the dumb presentational component to see if it got the correct props, i.e. the props data is congruent with what was dispatched to the store. here is an [example](https://github.com/onaio/reveal-frontend/blob/90a710b0497110a1e63c939330b050ca8f298645/src/containers/pages/PractitionerViews/PractitionerListView/tests/index.test.tsx#L104)

clients/core/src/containers/Households/List/tests/index.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+import toJson from 'enzyme-to-json';
+import * as React from 'react';
+import { Provider } from 'react-redux';
+import store from '../../../../store';
+import householdsReducer, { reducerName as householdsReducerName } from '../../../../store/ducks/households';
+import { HouseholdList, HouseholdListProps } from '../index';
+
+reducerRegistry.register(householdsReducerName, householdsReducer);
+
+jest.mock('../../../../configs/env');
```
p-netm: You might want to add the variables that are targeted by this mock to the `configs/__mocks__/env.ts` file, refer [here](https://jestjs.io/docs/en/manual-mocks#mocking-user-modules)

clients/core/src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+        removeHouseholdsActionCreator();
+        this.setState(
+            {
+                ...this.state,
+                loading: false,
+            },
+            () => {
+                fetchHouseholdsActionCreator(response.clients);
+                setTotalRecordsActionCreator(response.total);
+            },
```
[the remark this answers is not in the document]
prodhan29: put those serviceCall in a `then` block and also add a catch block

clients/core/src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+    pageNumber: string;
+    searchPlaceholder: string;
+    clientType: string;
+}
+
+export const defaultHouseholdListState: HouseholdListState = {
+    loading: true,
+    pageNumber: '1',
+    searchPlaceholder: 'Search Household',
+    clientType: 'ec_family',
```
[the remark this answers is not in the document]
prodhan29: Add a new constant variable `HOUSEHOLD_CLIENT_TYPE` in a constants.tsx file and move the `ec_family` there. Server accepts both string and number type variable. so changed the pagNumber data type from string to number.

clients/core/src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+                </div>
+            );
+        }
+    }
+
+    private search = (searchText: string) => {
+        this.getHouseholdList({ searchText });
+    };
+
+    private async getHouseholdList(extraParams: FlexObject = {}) {
```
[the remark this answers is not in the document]
prodhan29: this extraParam sometimes contain `searchText` or `level` or both. in current state there is no `level` variable but i saw it in the mock UI.

clients/core/src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+        const clientType = this.state.clientType;
+        const params = {
+            clientType,
+            pageNumber: this.state.pageNumber,
```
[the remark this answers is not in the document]
prodhan29: currently, there is no pager component in the app. In last meeting we were told it will be provided to us very soon. so changing state of the pageNumber will be seen in the code when it will be added

clients/core/src/containers/Households/List/tests/index.test.tsx
prodhan29: Added a test case like an example you provided.

clients/core/src/containers/Households/List/index.tsx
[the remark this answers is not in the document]
p-netm: ok

clients/core/src/containers/Households/List/tests/index.test.tsx
```diff
@@ -53,13 +55,70 @@ describe('containers/households/list/Householdlist', () => {
...
+            totalRecordsCount: 0,
+        };
+        const wrapper = mount(<HouseholdList {...props} />);
+
+        // initially the loading icon would show
+        const loadingComponent = wrapper.find('.lds-ripple-parent');
+        expect(loadingComponent.length).toBe(1);
+
+        // showing household-list container
+        wrapper.setState({ ...wrapper.state(), loading: false });
```
p-netm: instead of explicitely modifying the state you could force resolve the pending  promise like :
```typescript
it('render ...',  async () => {
....
expect(loadingComponent.length).toBe(1)

await new Promise(r => setImmediate(resolve))
... // continue with checking the rendered nodes
})```

Comment (p-netm) [reply clients/core/src/containers/Households/List/tests/index.test.tsx]: @prodhan29 
- Should the coomponent show the loader?, if yes, then we don't need to modify state like this, however if we should not be rendering the loader, you need to find out what is causing the loading state to remain true, we really should try testing code in a way users would interact with the component

**Comment by p-netm** [comment]
to avoid this error/ warning:
```sh
 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
        in HouseholdList
        in Provider (created by WrapperComponent)
        in WrapperComponent

```

Make sure to flush all pending promises before unmounting where a promise was involved in rendering the component

sarfaraz0075:
For Nahid to start working on the issue #204, this PR needs to be merged. @hmakandi and @p-netm, please review and if OK, merge this as soon as possible.

clients/core/src/configs/env.ts
```diff
@@ -112,3 +112,10 @@ export type OPENSRP_API_BASE_URL = typeof OPENSRP_API_BASE_URL;
 /** the clients endpoint NOTE: does not end with / */
 export const OPENSRP_CLIENT_ENDPOINT = process.env.REACT_APP_OPENSRP_CLIENT_ENDPOINT || 'client/search';
 export type OPENSRP_CLIENT_ENDPOINT = typeof OPENSRP_CLIENT_ENDPOINT;
+
+/** the households endpoint NOTE: does not end with / */
+export const OPENSRP_HOUSEHOLD_ENDPOINT = process.env.REACT_APP_OPENSRP_HOUSEHOLD_ENDPOINT || 'client/searchByCriteria';
```
p-netm: lets move this to constants.

clients/core/src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+                                                <td>{household.dateCreated}</td>
+                                                <td className="members-table-field">
+                                                    {household.attributes.dynamicProperties.member_count}
+                                                </td>
+                                                <td>
+                                                    <Link to={`/household/profile/${household.baseEntityId}/`}>
+                                                        View
+                                                    </Link>
+                                                </td>
+                                            </tr>
```
p-netm: Should we not replace the table here.

clients/core/src/containers/Households/List/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,246 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
p-netm: is it possible, to reduce this snapshot, can we do so?

p-netm:
@prodhan29 It would also be really nice if you added a snapshot of the page.

prodhan29:
> @prodhan29 It would also be really nice if you added a snapshot of the page.

Screenshot from 2020-03-24 20-50-20

clients/core/src/configs/env.ts
prodhan29: done

clients/core/src/containers/Households/List/tests/__snapshots__/index.test.tsx.snap
prodhan29: it creates a circular json thats why it gets bigger.. we can match a node instead of using the snapshot.

prodhan29: @p-netm update the test code to acheive the same result

clients/core/src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+        console.log('total records', this.props.totalRecordsCount);
+        return {
+            onPageChangeHandler: this.onPageChange,
+            pageNeighbors: PAGINATION_NEIGBOURS,
+            pageSize: PAGINATION_SIZE,
+            totalRecords: this.props.totalRecordsCount,
+        };
+    };
+
+    public render(): ReturnType<React.FC> {
```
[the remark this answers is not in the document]
prodhan29: changed to React.ReactNode.

clients/core/src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+            },
+            () => {
+                this.getHouseholdList();
+            },
+        );
+    };
+
+    /** it returns the required options for pagination component */
+    getPaginationOptions = (): PaginationProps => {
+        console.log('total records', this.props.totalRecordsCount);
```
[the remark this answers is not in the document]
prodhan29: Done !

clients/core/src/containers/Households/List/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+            );
+        }
+    }
+
+    private search = (searchText: string): void => {
+        this.getHouseholdList({ searchText });
+    };
+
+    /**
+     * it can contain searchText or level or both
```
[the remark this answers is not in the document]
prodhan29: Done !

prodhan29:
@p-netm looks like pagination is working. please review this branch as well.

PR #147, opened 2020-01-01
Add household-profile

prodhan29:
- Add Household-Profile component
- Test case for household-Profile

### DIFF
--- a/clients/core/src/containers/Households/Profile/index.tsx
+++ b/clients/core/src/containers/Households/Profile/index.tsx
@@ -1,0 +1,226 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import reducerRegistry from '@onaio/redux-reducer-registry';
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { Link, RouteComponentProps, withRouter } from 'react-router-dom';
+import { Col, Container, Row, Table } from 'reactstrap';
+import { Store } from 'redux';
+import Loading from '../../../components/page/Loading';
+import {
+    OPENSRP_CLIENT_ENDPOINT,
+    OPENSRP_EVENT_ENDPOINT,
+    OPENSRP_HOUSEHOLD_ENDPOINT,
+} from '../../../configs/env';
+import { HOUSEHOLD_URL } from '../../../constants';
+import { OpenSRPService } from '../../../services/opensrp';
+import clientReducer, {
+    Client,
+    fetchClients,
+    getClientsArray,
+    reducerName as clientsReducer,
+    removeClients,
+} from '../../../store/ducks/clients';
+import eventReducer, {
+    Event,
+    fetchEvents,
+    getEventsArray,
+    reducerName as eventReducerName,
+} from '../../../store/ducks/events';
+import householdsReducer, {
+    fetchHouseholds,
+    getHouseholdsArray,
+    Household,
+    reducerName as householdReducerName,
+} from '../../../store/ducks/households';
+import './householdProfile.css';
+
+/** register the event reducer */
+reducerRegistry.register(eventReducerName, eventReducer);
+
+/** register the client reducer */
+reducerRegistry.register(clientsReducer, clientReducer);
+
+/** register the household reducer */
+reducerRegistry.register(householdReducerName, householdsReducer);
+
+/** interface for HouseholdProfile URL params */
+interface HouseholdProfileURLParams {
+    id: string;
+}
+
+/** interface for HouseholdProfileProps */
+export interface HouseholdProfileProps extends RouteComponentProps<HouseholdProfileURLParams> {
+    household: Household | null;
+    events: Event[];
+    members: Client[];
+    fetchClientActionCreator: typeof fetchHouseholds;
+    fetchMembersActionCreator: typeof fetchClients;
+    fetchEventsActionCreator: typeof fetchEvents;
+    removeMembersActionCreator: typeof removeClients;
+    opensrpService: typeof OpenSRPService;
+}
+
+class HouseholdProfile extends React.Component<HouseholdProfileProps> {
+    public async componentDidMount() {
+        const {
+            fetchClientActionCreator,
+            fetchMembersActionCreator,
+            fetchEventsActionCreator,
+            match,
+            removeMembersActionCreator,
+        } = this.props;
+        const householdId = match.params.id || '';
+        const params = { identifier: householdId };
+        const clientService = new OpenSRPService(`${OPENSRP_CLIENT_ENDPOINT}`);
+        const clientResponse = await clientService.list(params);
+        if (clientResponse[0]) {
+            fetchClientActionCreator(clientResponse);
+            const eventService = new OpenSRPService(`${OPENSRP_EVENT_ENDPOINT}`);
+            const eventsResponse = await eventService.list(params);
+            fetchEventsActionCreator(eventsResponse);
+            const memberParams = {
+                baseEntityId: householdId,
+                clientType: 'householdMember',
+            };
+            const memberService = new OpenSRPService(`${OPENSRP_HOUSEHOLD_ENDPOINT}`);
+            const membersResponse = await memberService.list(memberParams);
+            removeMembersActionCreator();
+            fetchMembersActionCreator(membersResponse.clients);
+        }
+    }
+    public render() {
+        const { household, events, members } = this.props;
+        if (!household) {
+            return <Loading />;
+        }
+        return (
+            <Container id="householdProfile">
+                <div className="page-title">
+                    <span className="back-btn-bg">
+                        <Link to={`${HOUSEHOLD_URL}`}>
+                            <FontAwesomeIcon icon="arrow-left" />
+                            <span className="back-btn"> Back to Household </span>
+                        </Link>
+                    </span>
+                    <h3> {household.lastName} </h3>
+                </div>
+                <div id="basic-info-container">
+                    <Row className="basic-info-header-bg">
+                        <Col className="basic-info-header">
+                            <span className="basic-info-title"> Basic Information</span>
+                            <div className="float-right basic-info-edit">
+                                <a href={`${'#'}`}>Edit Profile</a>
+                            </div>
+                        </Col>
+                    </Row>
+                    <Row>
+                        <Col className="basic-info-body">
+                            <Table className="basic-info-table" borderless={true}>
+                                <tbody>
+                                    <tr>
+                                        <td className="basic-info-label">HHID Number</td>
+                                        <td>{household.baseEntityId}</td>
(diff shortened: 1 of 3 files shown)


clients/core/src/containers/Households/Profile/helpers/tableDefinition.tsx
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable react/prop-types */
+import React, { useMemo } from 'react';
+import { Link } from 'react-router-dom';
+import { VIEW } from '../../../../constants';
+import { UseTableCellProps } from 'react-table';
+import { Client } from '../../../../store/ducks/clients';
+// Child specific configurable table columns
```
p-netm, line 7: lets remove this, seems like a stowaway docstring

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+    location: createLocation(matchVariable.url),
+    match: matchVariable,
+};
+
+class HouseholdProfile extends React.Component<HouseholdProfileProps> {
+    public static defaultProps: HouseholdProfileProps = defaultProfileProps;
+
+    public async componentDidMount(): Promise<void> {
+        const { fetchClient, fetchMembers, fetchEvents, match, removeMembers, opensrpService } = this.props;
+        const householdId = match.params.id || 'ea0edc48-4752-4ad0-a834-f1f68c7ae310';
```
p-netm, line 109: Do we have to provide a default value for household id like this? do you think it would be better to only execute the api call when there is a valid household id?

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+            const eventsResponse = await eventService.list(params);
+            fetchEvents(eventsResponse);
+            const memberParams = {
+                baseEntityId: householdId,
+                clientType: 'householdMember',
+            };
+            const memberService = new opensrpService(OPENSRP_API_BASE_URL, OPENSRP_CLIENT_ENDPOINT, generateOptions);
+            const membersResponse = await memberService.list(memberParams);
+            removeMembers();
+            fetchMembers(membersResponse.clients);
```
p-netm, line 125: can we also abstract this into a function and also it seems like it would lead for cleaner code if we were to chain the promise callbacks. ie. `Promise(resolve => resolve('stuff')).then(response => // use response).catch(err => // do something with error)`

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+        const { household, events, members } = this.props;
+        if (!household) {
+            return <Loading />;
+        }
+        return (
+            <Container id="household-profile">
+                <div className="page-title">
+                    <span className="back-btn-bg">
+                        <Link to={`${HOUSEHOLD_URL}`}>
+                            <FontAwesomeIcon icon="arrow-left" />
```
p-netm, line 138: lets add this to the library. https://www.npmjs.com/package/@fortawesome/react-fontawesome#build-a-library-to-reference-icons-throughout-your-app-more-conveniently

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+    household: Household | null;
+    events: Event[];
+    members: Household[];
+}
+
+/** Map props to state  */
+const mapStateToProps = (state: Partial<Store>): DispatchedStateProps => {
+    const result = {
+        events: getEventsArray(state),
+        household: getHouseholdsArray(state).length > 0 ? getHouseholdsArray(state)[0] : null,
```
p-netm, line 198: I think we should use `getHouseholdById` here

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+    return result;
+};
+
+/** map props to actions */
+const mapDispatchToProps = {
+    fetchClient: fetchHouseholds,
+    fetchEvents: fetchEvents,
+    fetchMembers: fetchClients,
+    removeMembers: removeClients,
+};
```
p-netm, line 210: I think the naming here can be made better, especially to avoid confusion, say here it seems like it makes more sense that the `fetchClient` prop would have the value `fetchClients` other than the `fetchHouseholds`

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+                                </tr>
+                            </tbody>
+                            <tbody>
+                                <tr>
+                                    <td className="info-label">Head of Household</td>
+                                    <td>{household.firstName}</td>
+                                    <td className="info-label">Register date</td>
+                                    <td>{household.dateCreated || ''}</td>
+                                </tr>
+                            </tbody>
```
p-netm, line 170: lets move this section into the infoCard section. the goal would be to make it generic such that you can feed the component an object and it should be able to render the information with an ui that is similar to the mockups so in this case it would be the `basic information` body section.

clients/core/src/containers/Households/Profile/helpers/tableDefinition.tsx
```diff
@@ -0,0 +1,58 @@
...
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export const useMemberTableColumns = () => {
+    return useMemo(
+        () => [
+            {
+                /** placeholder for location table crumb */
+                // TODO - the location tableCrumb will be a render prop
+                // eslint-disable-next-line react/display-name
+                Header: (): React.ReactNode => <></>,
```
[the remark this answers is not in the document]
prodhan29: @p-netm the example you have provided, when i apply this, it doesn't show the column name, i have also seen the examples of react-table given in its official document, those are exactly the same as yours. but after applying this unfortunately, column name goes missing. when i left the header property with empty string, the table does show the column names.

clients/core/src/containers/Households/Profile/helpers/tableDefinition.tsx
prodhan29: removed

clients/core/src/containers/Households/Profile/helpers/tableDefinition.tsx
```diff
@@ -0,0 +1,58 @@
...
+                    },
+
+                    {
+                        Header: 'Register Status',
+                        accessor: 'attributes.dynamicProperties.registration_status',
+                        disableSortBy: true,
+                    },
+                    {
+                        // eslint-disable-next-line react/display-name
+                        Cell: ({ row: { values } }: UseTableCellProps<Client>): ReturnType<React.FC> => (
```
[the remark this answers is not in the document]
prodhan29: changed the return type to React.ReactNode

clients/core/src/containers/Households/Profile/index.tsx
prodhan29: removed that. used that default id to test the api call, as i have no household-list to directly come to the profile component.

clients/core/src/containers/Households/Profile/index.tsx
prodhan29: refactored the server call.

clients/core/src/containers/Households/Profile/index.tsx
prodhan29: Done !

clients/core/src/containers/Households/Profile/index.tsx
prodhan29: renamed the property

clients/core/src/containers/Households/Profile/index.tsx
prodhan29: @p-netm @redoy11  as far i understand, infocard is like a container . where header is inline and body will be a reuseable component through composition,  body could be basicInfo, current-register or growth-chart.  here I used composition to render the basicInfo...
if i move this section into the infocard, i dont know how i will render current-register or growth chart later.  
what i can do is to create another generic component `basicInfo`, i will feed this component an object that will render the ui given in mockup and then render the `infocard` with `basicinfo` through composition.

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+export interface HouseholdProfileProps extends RouteComponentProps<HouseholdProfileURLParams> {
+    household: Household | null;
+    events: Event[];
+    members: Client[];
+    fetchClient: typeof fetchHouseholds;
+    fetchMembers: typeof fetchClients;
+    fetchEvents: typeof fetchEvents;
+    removeMembers: typeof removeClients;
+    opensrpService: typeof OpenSRPService;
+}
```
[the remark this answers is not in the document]
prodhan29: Updated the porptype as you suggested

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+const matchVariable: match<{ id: string }> = {
+    isExact: false,
+    params: { id: '1' },
+    path: '/household/profile/:id/',
+    url: `/household/profile/${1}/`,
+};
```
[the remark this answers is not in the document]
prodhan29: removed this matchVariable, history, match

clients/core/src/containers/Households/Profile/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+    events: [],
+    members: [],
+    fetchClient: fetchHouseholds,
+    fetchMembers: fetchClients,
+    fetchEvents: fetchEvents,
+    removeMembers: removeClients,
+    opensrpService: OpenSRPService,
+    history: createMemoryHistory(),
+    location: createLocation(matchVariable.url),
+    match: matchVariable,
```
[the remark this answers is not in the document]
prodhan29: removed this matchVariable, history, match

clients/core/src/containers/Households/Profile/index.tsx
p-netm: lets create another component and have composition take place there. so like create a basicInfo componnt and have it have the body as shown here, only that it should be generic enough to use any object as input data, and render.

clients/core/src/containers/Households/Profile/helpers/tableDefinition.tsx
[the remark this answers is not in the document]
p-netm: @prodhan29 https://codesandbox.io/s/sample-opensrp-headers-5u8sf?file=/src/table.js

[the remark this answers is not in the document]
prodhan29: there was a css which is actually hiding the header of the table. after removing that the table is now showing the headers

Issue #108, opened 2019-11-19
Compartment circles should be clikable


oluwatobim:
Only the numbers are clickable (29) but as seen on the left red circle, the whole circle should be clickable.
![image](https://user-images.githubusercontent.com/57088835/69163021-d4169c80-0aed-11ea-8b7b-167141b53c3f.png)

hmakandi:
This is implemented and confirmed

PR #143, opened 2019-12-23
Issue#108

FrankApiyo:
Fixes: #108

PR #153, opened 2020-01-08
Make compartiment circles clickable

FrankApiyo:
Fixes: #108

### DIFF
--- a/src/components/DataCircleCard/index.tsx
+++ b/src/components/DataCircleCard/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { Link } from 'react-router-dom';
+import { Link, RouteComponentProps, withRouter } from 'react-router-dom';
 import { Card, CardBody, CardTitle } from 'reactstrap';
 import {
   ALL,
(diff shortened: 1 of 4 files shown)


FrankApiyo:
This PR replaces this https://github.com/OpenSRP/opensrp-web/pull/143

FrankApiyo:
[This](https://github.com/OpenSRP/opensrp-web/pull/143#pullrequestreview-337767416l) comment has also been responded to here

src/components/DataCircleCard/index.tsx
```diff
@@ -138,24 +139,26 @@ function DataCircleCard({
...
-                  // tslint:disable-next-line: jsx-no-lambda
-                  onClick={() => {
-                    if (filterArgs) {
-                      addFilterArgsActionCreator(filterArgs);
-                    }
-                  }}
-                >
-                  <span className="number">{spec.riskValue}</span>
-                </Link>
+              <li
```
moshthepitt: Why are opting out of using `Link`?

FrankApiyo: There really wasn't a good reason for this decision except that the route taken here was the easier route.

@moshthepitt  do you think we should stick with `<Link>` ?

moshthepitt: Yes I think we should stick with Link if possible

FrankApiyo: @moshthepitt after replacing the `<li>` with `<Link>`, I noticed how this makes the code more declarative as opposed to imperative philosophy of react.

PR #171, opened 2020-01-15
Add Child dux module

prodhan29:
- add dux file of childs.
- add test code for the child-dux module

### DIFF
--- a/clients/core/src/store/ducks/childs.ts
+++ b/clients/core/src/store/ducks/childs.ts
@@ -76,3 +76,61 @@ const initialState: ImmutableChildState = SeamlessImmutable({
     childsById: {},
     totalRecords: 0,
 });
+
+/** the child reducer function */
+export default function reducer(
+    state: ImmutableChildState = initialState,
+    action: ChildActionTypes,
+): ImmutableChildState {
+    switch (action.type) {
+        case CHILDS_FETCHED:
+            return SeamlessImmutable({
+                ...state,
+                childsById: { ...state.childsById, ...action.childsById },
+            });
+        case REMOVE_CHILDS:
+            return SeamlessImmutable({
+                ...state,
+                childsById: action.childsById,
+            });
+        case SET_TOTAL_RECORDS:
+            return SeamlessImmutable({
+                ...state,
+                totalRecords: action.totalRecords,
+            });
+        default:
+            return state;
+    }
+}
+
+/** returns all child in the store as values whose keys are their respective ids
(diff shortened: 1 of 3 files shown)


moshthepitt:
Is there a way to avoid repeating the code?

prodhan29:
> Is there a way to avoid repeating the code?

these are the basic functionalities which are common to all dux module.  in the mock ui, i saw some filtering criteria different from each others. however @p-netm  requested to create separate reducers for each page. you may suggest some way to avoid improve

p-netm:
> Is there a way to avoid repeating the code?

@moshthepitt @prodhan29 

I fear that further abstraction in an attempt to dry the code out, will result in a loss of semantics around the data handled by any of the dux modules, its is however still possible to do so.

moshthepitt:
@p-netm @prodhan29 

Based on the premise that usually being DRY is better than not being DRY, I have some questions:

- Do you get all these households, clients, and children from the same OpenSRP endpoint?
- If you consider [households](https://github.com/OpenSRP/opensrp-web/blob/master/clients/core/src/store/ducks/clients.ts) and [clients](https://github.com/OpenSRP/opensrp-web/blob/master/clients/core/src/store/ducks/households.ts):
    - Is it possible to make the reducer function more DRY?  The action types are different but what you do after identifying them is identical
    - is it possible to make the selectors that do the same thing more DRY?
    - is it possible to use generic types to improve the DRYness of the interfaces and other types? I mean in the case of the actions you are already extending AnyAction, why not go further?

p-netm:
>     * Do you get all these households, clients, and children from the same OpenSRP endpoint?

Yes, i did consider using a single reducer, the tradeoffs in terms of efficiency would be beteen speed and duplication of data, whereby having a single reducer would require filtering through thousands of records just to find say all clients of client_type child, or ANC, having different reducers for each would solve this but then getting data for households members would result in data duplication solely in the ducks slice containing household's members data.

>     * Is it possible to make the reducer function more DRY?  The action types are different but what you do after identifying them is identical

Yes, @prodhan29 ?

>     * is it possible to make the selectors that do the same thing more DRY?

yes, @prodhan29 ?

prodhan29:
> 
> > ```
> > * Is it possible to make the reducer function more DRY?  The action types are different but what you do after identifying them is identical
> > ```
> 
> Yes, @prodhan29 ?
> 
> > ```
> > * is it possible to make the selectors that do the same thing more DRY?
> > ```
> 
> yes, @prodhan29 ?
i see both approach has their own trade off. please decide which would be better

PR #175, opened 2020-01-15
Show a message if theres no record for a commune

### LINKED ISSUE #174: At the end of the locations disaggregation(All the way to the commune), if there is no patient to show in the log the village level, display there is no patient in this record.
Part of Issue #154

FrankApiyo:
Fixes #174

### DIFF
--- a/src/components/NoRecord/index.tsx
+++ b/src/components/NoRecord/index.tsx
@@ -1,0 +1,5 @@
+import React from 'react';
+
+export default function NoRecord() {
+  return <div>no record</div>;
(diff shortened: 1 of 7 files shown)


src/components/NoRecord/index.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
moshthepitt: Please add docstrings here.

FrankApiyo: done!

PR #178, opened 2020-01-16
Pagination

p-netm:
- Will add a stateful functional component that we can use for pagination.


moshthepitt:
@p-netm curious why this is not a package?  It looks somewhat generic...

clients/core/package.json
```diff
@@ -12,6 +12,7 @@
         "@onaio/connected-reducer-registry": "^0.0.1",
         "@onaio/gatekeeper": "^0.0.14",
         "@onaio/loaders": "^0.0.1",
+        "@onaio/pagination": "^0.0.3",
```
cKellyDesign: Are we using this anywhere?

yarn.lock
```diff
@@ -9768,56 +9776,6 @@ opencollective-postinstall@^2.0.2:
...
-    react-router "^5.0.1"
-    react-router-dom "^5.0.1"
-    react-scripts "^3.0.1"
-    reactstrap "^8.0.1"
-    redux "^4.0.4"
-    redux-thunk "^2.3.0"
-    seamless-immutable "^7.1.4"
-    snapshot-diff "^0.5.2"
-    typescript "^3.5.3"
-
```
cKellyDesign: Why does this get dropped?

p-netm: My bad,  I will fix this, Thanks.

clients/core/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,206 @@
+/* eslint-disable react/prop-types */
+import * as React from 'react';
+import { ActionTypes, PaginationOptions, PaginationState, usePagination } from '@onaio/pagination';
```
p-netm: @cKellyDesign This is where `@onaio/pagination` is primarily used.

p-netm:
> @p-netm curious why this is not a package? It looks somewhat generic...


> Agreed. Is there a reason the changes in this PR are not a `v0.0.4` of the `@onaio/pagination` package?

The pagination in `onaio/jstools` contains the minimum logic required to create any sort of generic pagination. This component on the other hand adds a few things that are custom to opensrp-web
- The JSX file adds custom html that works renders a bootstrap styled component
- The index file adds extra logic that makes it possible for `@onaio/jstools` support the bootstrap component 
- Also adds functionality that enables the pagination component get more data from an API -> This is not usual functionality for a pagination component

I do not have any reservations to adding this to the `@onaio/jstools`, just wanted to point out the above things.

moshthepitt:
@p-netm cool, thanks.  Please add it as a package - it seems to me that at least the bootstrap things may be reused a lot.

p-netm:
@moshthepitt @cKellyDesign I have created an issue to move [this](https://github.com/onaio/js-tools/issues/135), i hope this is acceptable for now.

PR #179, opened 2020-01-16
Add location dux module

prodhan29:
- add ducks module location-based-breadcrumb 
- test code for duck module

### DIFF
--- a/clients/core/src/store/ducks/locations.ts
+++ b/clients/core/src/store/ducks/locations.ts
@@ -4,29 +4,32 @@ import SeamlessImmutable from 'seamless-immutable';
 /** The reducer name */
 export const reducerName = 'locations';
 
-/** A single node for location hierarchy */
-export interface LocationNode {
-    [key: string]: {
-        id: string;
-        label: string;
-        node: {
+/** Inteface for Location */
+export interface Location {
+    id: string;
+    label: string;
+    node: {
+        locationId: string;
+        name: string;
+        parentLocation?: {
             locationId: string;
             name: string;
+            voided: boolean;
             parentLocation?: {
                 locationId: string;
                 name: string;
                 voided: boolean;
-                parentLocation?: {
-                    locationId: string;
-                    name: string;
-                    voided: boolean;
-                };
             };
-            tags: string[];
-            voided: boolean;
         };
-        children?: LocationNode;
+        tags: string[];
+        voided: boolean;
     };
+    children?: LocationNode;
+}
+
+/** A single node for location hierarchy */
+export interface LocationNode {
+    [key: string]: Location;
 }
 
 /** Interface for location object */
--- a/clients/core/src/store/ducks/locations.ts
+++ b/clients/core/src/store/ducks/locations.ts
@@ -54,7 +54,7 @@ interface SetLocationAction extends AnyAction {
 export type LocationActionTypes = SetLocationAction | AnyAction;
 
 /** setLocation action */
-export const setLocations = (locations: LocationHierarchy): SetLocationAction => ({
+export const setLocations = (locations: LocationHierarchy | null): SetLocationAction => ({
     locations,
     type: SET_LOCATIONS,
 });
--- a/clients/core/src/store/ducks/locations.ts
+++ b/clients/core/src/store/ducks/locations.ts
@@ -1,0 +1,55 @@
+import { AnyAction, Store } from 'redux';
+import SeamlessImmutable from 'seamless-immutable';
+
+/** The reducer name */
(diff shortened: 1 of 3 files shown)


clients/core/src/store/ducks/locations.ts
```diff
@@ -0,0 +1,85 @@
+import { AnyAction, Store } from 'redux';
+import SeamlessImmutable from 'seamless-immutable';
+
+/** The reducer name */
+export const reducerName = 'locations';
+
+/** A single node for location hierarchy */
+export interface LocationNode {
```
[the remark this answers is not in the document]
prodhan29: yes. here is an example. [https://github.com/prodhan29/breadcrum-location/blob/real-location-data/src/sample-data.js](url)

clients/core/src/store/ducks/tests/location.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        store.dispatch(setLocations(fixtures.location));
+        expect(getLocationHierarchy(store.getState())).toEqual(fixtures.location);
+    });
+
+    it(' spy on location hierarchy', () => {
+        const spyOnLocationHierarchy = jest.spyOn(locationDucksModule, 'setLocations');
+        store.dispatch(setLocations(fixtures.location));
+        expect(spyOnLocationHierarchy).toHaveBeenCalledTimes(1);
+        expect(spyOnLocationHierarchy).toHaveBeenCalledWith(fixtures.location);
+    });
```
p-netm: What exactly is being tested here?

clients/core/src/store/ducks/tests/fixtures.ts
```diff
@@ -335,5 +336,39 @@ export const household3: Household = {
     _rev: 'v1',
 };
 
+export const location: LocationHierarchy = {
```
p-netm: can we include a few more fixtures for this, or a more comprehensive one

clients/core/src/store/ducks/locations.ts
```diff
@@ -0,0 +1,96 @@
...
+export const reducerName = 'locations';
+
+/** Inteface for Location */
+export interface Location {
+    id: string;
+    label: string;
+    node: {
+        locationId: string;
+        name: string;
+        parentLocation?: {
```
p-netm: Breaking down this interface could make it easier to use recursion in declaring this interface. for instance: 
```typescript
interface Parent {
  name: string;
  child?: Parent
}

const matryoshka: Parent = {
  name: 'grandparents',
  child: {
    name: 'parents',
    child: {
      name: 'child',
    }
  }
}
```

clients/core/src/store/ducks/tests/fixtures.ts
prodhan29: make it more comprehensive

clients/core/src/store/ducks/locations.ts
prodhan29: Done.

clients/core/src/store/ducks/tests/location.test.ts
prodhan29: making sure that action is called not more than once.

clients/core/src/store/ducks/tests/fixtures.ts
prodhan29: update the location fixture.

clients/core/src/store/ducks/tests/location.test.ts
p-netm: my problem with this test is that, `setLocations` could be any function, doing God knows what, but still this test would pass.

p-netm: it might seem unlikely to occur in production, but maybe also add a test that sets locations on a non-empty state.

clients/core/src/store/ducks/tests/location.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        expect(getLocationHierarchy(store.getState())).toEqual({});
+    });
+
+    it(' set locations correctly', () => {
+        store.dispatch(setLocations(fixtures.location));
+        expect(getLocationHierarchy(store.getState())).toEqual(fixtures.location);
+    });
+
+    it(' can set locations on a non empty state', () => {
+        store.dispatch(setLocations(null));
```
p-netm: instead of using null add one more fixture for this.

prodhan29: okay

prodhan29: @p-netm done

PR #186, opened 2020-01-23
Turn all eslint errors into warnings

p-netm:
Why: It really irks me when linting issues are "scriggly-lined" as errors, especially given the fact that they are automaticaly fixed later.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,7 @@ module.exports =  {
   rules:  {
     '@typescript-eslint/explicit-function-return-type': 'warn',
     '@typescript-eslint/no-inferrable-types': 'warn',
+    "prettier/prettier": "warn",
   },
   settings:  {
     react:  {


moshthepitt:
Thanks for bringing this up.  I have been thinking of linting configs a lot lately.  I am currently heavily leaning towards linting rules to be as strict as possibly possible.  Here's some reading around this area

- [Strict linting](https://www.aaronbuxbaum.com/strict-linting/)
- [Squeezing maximum type safety out of Typescript with TypeStrict](https://blog.neufund.org/squeezing-maximum-type-safety-out-of-typescript-with-typestrict-382671c5326)
- [Typestrict](https://github.com/krzkaczor/TypeStrict)

We can talk about it some more, but if we are heading in that direction then we really should not relax any linting rules at all.

cc: @cKellyDesign to chime in with his thoughts

moshthepitt:
Going to close this.  Please reopen when ready to discuss further.

PR #194, opened 2020-02-13
refactor the existing dux module, add child dux and fix the test code

### LINKED ISSUE #189: Refactor the Client, Household, Child dux module using baseClient.ts
Currently, the dux modules related to the client are separated from each other and don't inherit the common operation from anywhere. but now we have an abstract file for all those common operations, need to use that as a parent file. clients, household, child register, all of their dux modules should be similar to the ANC module. Create PR for each of the dux modules for refactoring.

prodhan29:
This PR closes #189 and closes  #171 
The following things are present in the PR:
- refactor the client dux module, household dux module
- fix test code for client and household dux module 
- add child-dux module
- add test code for child-dux

### DIFF
--- a/clients/core/src/store/ducks/child.ts
+++ b/clients/core/src/store/ducks/child.ts
@@ -11,23 +11,23 @@ import {
 import { Store } from 'redux';
 
 /** The reducer name */
-export const reducerName = 'opensrp-web/client-type/households';
+export const reducerName = 'opensrp-web/client-type/child';
 
 /** Interface for household object same as client */
-export type Household = Client;
+export type Child = Client;
 
 /** Client Reducer */
-const reducer = reducerFactory<Household>(reducerName);
+const reducer = reducerFactory<Child>(reducerName);
 
 // action creators
-export const fetchHouseholds = fetchClientsFactory<Household>(reducerName);
-export const removeHouseholds = removeClientsFactory(reducerName);
+export const fetchChildList = fetchClientsFactory<Child>(reducerName);
+export const removeChildList = removeClientsFactory(reducerName);
 export const setTotalRecords = setTotalRecordsFactory(reducerName);
 
 // selectors
-export const getHouseholdsById = getClientsByIdFactory<Household>(reducerName);
-export const getHouseholdsArray = getClientsArrayFactory<Household>(reducerName);
-export const getHouseholdById = getClientByIdFactory<Household>(reducerName);
+export const getChildListById = getClientsByIdFactory<Child>(reducerName);
+export const getChildArray = getClientsArrayFactory<Child>(reducerName);
+export const getChildById = getClientByIdFactory<Child>(reducerName);
 
 export function getTotalRecords(state: Partial<Store>): number {
     return (state as any)[reducerName].totalRecords;
--- a/clients/core/src/store/ducks/child.ts
+++ b/clients/core/src/store/ducks/child.ts
@@ -7,13 +7,13 @@ import {
     getClientsByIdFactory,
     getClientsArrayFactory,
     getClientByIdFactory,
+    getTotalRecordsFactory,
 } from './baseClients';
-import { Store } from 'redux';
 
 /** The reducer name */
 export const reducerName = 'opensrp-web/client-type/child';
 
-/** Interface for household object same as client */
+/** Interface for Child object same as client */
 export type Child = Client;
 
 /** Client Reducer */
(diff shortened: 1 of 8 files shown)


clients/core/src/store/ducks/child.ts
```diff
@@ -0,0 +1,36 @@
...
+export const fetchChildList = fetchClientsFactory<Child>(reducerName);
+export const removeChildList = removeClientsFactory(reducerName);
+export const setTotalRecords = setTotalRecordsFactory(reducerName);
+
+// selectors
+export const getChildListById = getClientsByIdFactory<Child>(reducerName);
+export const getChildArray = getClientsArrayFactory<Child>(reducerName);
+export const getChildById = getClientByIdFactory<Child>(reducerName);
+
+export function getTotalRecords(state: Partial<Store>): number {
```
p-netm: I believe there is also a factory for this. `getTotalRecordsFactory`

clients/core/src/store/ducks/households.ts
```diff
@@ -1,136 +1,36 @@
...
- */
-export function getHouseholdById(state: Partial<Store>, id: string): Household | null {
-    return get(getHouseholdsById(state), id) || null;
-}
-
-/** returns the count of all records present in server
- * @param {Partial<Store>} state - the redux store
- * @return { number } - total records value from the store
- */
 export function getTotalRecords(state: Partial<Store>): number {
```
p-netm: I believe there is also a factory for this. getTotalRecordsFactory

clients/core/src/store/ducks/clients.ts
```diff
@@ -31,114 +33,20 @@ export interface Client {
...
-}
-
-/** get a specific client by their id
- * @param {Partial<Store>} state - the redux store
- * @return {Client | null} a client obj if the id is found else null
- */
-export function getClientById(state: Partial<Store>, id: string): Client | null {
-    return get(getClientsById(state), id) || null;
-}
+export default reducer;
```
p-netm: do we also want to include the `TotalRecords` actions and selectors here?

clients/core/src/store/ducks/child.ts
```diff
@@ -0,0 +1,36 @@
...
+    getClientsByIdFactory,
+    getClientsArrayFactory,
+    getClientByIdFactory,
+} from './baseClients';
+import { Store } from 'redux';
+
+/** The reducer name */
+export const reducerName = 'opensrp-web/client-type/child';
+
+/** Interface for household object same as client */
```
p-netm: <del>household</del>

clients/core/src/store/ducks/tests/child.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    it('sets total records', () => {
+        store.dispatch(setTotalRecords(3));
+        let numberOfRecords = getTotalRecords(store.getState());
+        expect(numberOfRecords).toEqual(3);
+
+        store.dispatch(setTotalRecords(0));
+        numberOfRecords = getTotalRecords(store.getState());
+        expect(numberOfRecords).toEqual(0);
+    });
+});
```
p-netm: Nice
1 trivial test scenario that i think we are missing and can be a nice to have; is to to make sure dispatching the same object more than once adds  it only once to a specified store slice.

```typescript
    it('Adds new child-list to store instead of overwriting existing ones', () => {
        store.dispatch(fetchChildList([fixtures.child1, fixtures.child2]));
        let numberOfChild = getChildArray(store.getState()).length;
        expect(numberOfChild).toEqual(2);

        store.dispatch(fetchChildList([fixtures.child3]));
        numberOfChild = getChildArray(store.getState()).length;
        expect(numberOfChild).toEqual(3);
        
       store.dispatch(fetchChildList([fixtures.child1]));
        let numberOfChild = getChildArray(store.getState()).length;
        expect(numberOfChild).toEqual(3);
    });
```

I hope you don't mind adding this to the baseClients tests as well,

clients/core/src/store/ducks/clients.ts
```diff
@@ -31,114 +33,20 @@ export interface Client {
...
-            return SeamlessImmutable({
-                ...state,
-                clientsById: action.clientsById,
-            });
-        default:
-            return state;
-    }
-}
+// action
+/** actionCreator returns action to to add anc records to store */
```
p-netm: Tiny cleanup here in the docstring <del> anc <del>

clients/core/src/store/ducks/child.ts
prodhan29: didn't notice that at first. updated the code with `getTotalRecordsFactory`

clients/core/src/store/ducks/child.ts
prodhan29: done

clients/core/src/store/ducks/tests/child.test.ts
prodhan29: Added a test case in baseClient. to make sure same object won't get saved twice.

clients/core/src/store/ducks/clients.ts
prodhan29: done.

PR #198, opened 2020-02-18
Client profile view

prodhan29:
- add client profile view
- add test code for client-profile

### DIFF
--- a/clients/core/src/containers/Clients/Profile/index.tsx
+++ b/clients/core/src/containers/Clients/Profile/index.tsx
@@ -1,0 +1,165 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { Link, RouteComponentProps, withRouter } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


p-netm:
@prodhan29 

**InfoCard**

This will be base component for Basic information, household members, and the registers panels

- Make use of [render props](https://reactjs.org/docs/render-props.html) so that we can provide jsx or component that can be rendered as the header and/or the body

i.e. for Basic information : the card would comprise of 2 sections: the renderHeaderProp and the renderBodyProp

  - the header with the title `Basic Information` and the link to edit the profile. The header can either be declared inline or factored into its own component
  - The section would be a textual table that displays the clients details. This would be a different reusable component.

**The ancRegisters**
would have a component as the headerProp that has the  `CurrentRegisters and the quick links as tabs`

- The renderBodyProp is a pill like navigation component that will compromise of 2 components. A textual table component for client details, and a table that will be used to show the actual register event details.

p-netm:
@prodhan29 @redoy11  [This](https://github.com/OpenSRP/opensrp-web/pull/198#issuecomment-596183262) is  an initial draft and will soon finish up on the full thing. I however do not think the details above willl change and so you can go ahead and use that or the time being

PR #200, opened 2020-02-20
Add Dropdown component

prodhan29:
- contain the gender dropdown component
- test code for the component
- library used [react-select](https://react-select.com/home#getting-started)


clients/core/src/components/page/Dropdown/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    render() {
+        const { isSearchable } = this.state;
+        const { placeholder, dropdownOptions } = this.props;
+        return (
+            <div id="custom-dropdown">
+                <Select
+                    className="basic-single"
+                    classNamePrefix="select"
+                    onChange={(e: any) => this.props.selectCallBack(e.value)}
```
p-netm: do we have to use any here?

Issue #209, opened 2020-02-25
Search Box in Client Register Page


sarfaraz0075:
To complete the issue, the following needs to be done:

- Add Search by First Name 
- Add Search By Identifier
- Add related functionalities to Search By

Issue #208, opened 2020-02-25
Gender Filter in Client Register Page


sarfaraz0075:
To complete the issue, the following needs to be done:

- Add Filter By Gender
- Add related functionalities to Filter By.

PR #212, opened 2020-02-26
Filters in clients

prodhan29:
-following feature is in this pr
 

- searchText filter
- gender filter

it will resolve this issues: #208 #209


prodhan29:
@p-netm interface for the DropdownOption and default values for gender dropdown need to move some where else. please suggest a proper structure.

clients/core/src/containers/pages/Clients/List/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+    constructor(props: ClientListProps) {
+        super(props);
+        this.state = defaultClientListState;
+    }
+
+    public async componentDidMount() {
+        this.getDataFromServer();
+    }
+
+    getDataFromServer = async () => {
```
p-netm: lets add a docstring here, and any other functions that should have one but currently does not.

clients/core/src/containers/pages/Clients/List/index.tsx
```diff
@@ -37,6 +40,49 @@ const ClientTable: React.FC<ClientTableProps> = ({ tableData }: ClientTableProps
     <OpenSRPTable {...{ data: tableData, tableColumns: useClientTableColumns() }} />
 );
 
+/** interface for the pagination props */
+interface PaginationProps {
+    initialState: any;
```
prodhan29: @p-netm these are the required options that need to be passed. can you suggest what should be do except using `any` here

clients/core/src/configs/__mocks__/env.ts
```diff
@@ -8,7 +8,7 @@
 export const OPENSRP_API_BASE_URL = 'https://test.smartregister.org/opensrp/rest/';
 export type OPENSRP_API_BASE_URL = typeof OPENSRP_API_BASE_URL;
 /** the clients endpoint NOTE: does not end with / */
-export const OPENSRP_CLIENT_ENDPOINT = 'client/search';
+export const OPENSRP_CLIENT_ENDPOINT = 'client/searchByCriteria';
```
p-netm: can we move this to the constants file,

prodhan29:
@p-netm is this branch in a state to get merged with master ?, if yes, then please do it, so that i can apply those pagination changes to other branches. pagination issue still persist in [household-list](https://github.com/OpenSRP/opensrp-web/pull/146) branch

PR #213, opened 2020-02-26
Add a  custom reusable Grid component

p-netm:
- for the register pages, has support for sorting.
- Only needs to supply a list of columns just as the underlying react-table library requires, 
and data.

This is also part of the work that i had been doing from #128 , but after the rebase , i ended up with a very dirty diff.


clients/core/src/components/ReactTable/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+                                        </td>
+                                    );
+                                })}
+                            </tr>
+                        );
+                    })}
+                </tbody>
+            </table>
+            <br />
+            <div>Showing all rows</div>
```
cKellyDesign: As opposed to pagination?

clients/core/package.json
```diff
@@ -43,6 +43,7 @@
         "react-router": "^5.0.1",
         "react-router-dom": "^5.0.1",
         "react-scripts": "^3.0.1",
+        "react-table": "7.0.0-beta.15",
```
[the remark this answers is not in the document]
cKellyDesign: > Will it be a problem if...

Seems like it won't be as long as we keep the `v6` dependency in the package and don't try making `v6` tables with `v7` API.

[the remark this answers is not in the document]
p-netm: Yeah

clients/core/src/components/ReactTable/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+import './index.css';
+import { PLEASE_DEFINE_SOME_COLUMNS } from '../../constants';
+
+/** props for the ReactTable component */
+interface ReactTableProps<T extends object> {
+    /** a memoized list of columns  */
+    tableColumns: Column[];
+    /** the data to be rendered in table */
+    data: T[];
+}
```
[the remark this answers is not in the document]
p-netm: @cKellyDesign  My initial thoughts were that the data could be structured in whatever way, e.g. you can have a a multi dimensional list.

clients/core/src/components/ReactTable/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const defaultANCTableProps = {
```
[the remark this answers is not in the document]
p-netm: yes we shall.

clients/core/src/components/ReactTable/index.tsx
p-netm: will remove this.

clients/core/src/components/ReactTable/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+                                                )
+                                            ) : (
+                                                <>&nbsp;&nbsp;&nbsp;↕</>
+                                            )
+                                        ) : (
+                                            ''
+                                        )}
+                                    </span>
+                                </th>
+                            ))}
```
moshthepitt: I am getting a Typescript warning on these lines.  i,e,

```
(parameter) column: ColumnInstance<UseSortByColumnProps<{}>>
Argument of type '(column: ColumnInstance<UseSortByColumnProps<{}>>, index: number) => Element' is not assignable to parameter of type '(value: ColumnInstance<{}>, index: number, array: ColumnInstance<{}>[]) => Element'.
  Types of parameters 'column' and 'value' are incompatible.
    Type 'ColumnInstance<{}>' is not assignable to type 'ColumnInstance<UseSortByColumnProps<{}>>'.
      Types of property 'columns' are incompatible.
        Type 'Column<{}>[] | undefined' is not assignable to type 'Column<UseSortByColumnProps<{}>>[] | undefined'.
          Type 'Column<{}>[]' is not assignable to type 'Column<UseSortByColumnProps<{}>>[]'.
            Type 'Column<{}>' is not assignable to type 'Column<UseSortByColumnProps<{}>>'.
              Types of property 'show' are incompatible.
                Type 'boolean | ((instance: TableInstance<{}>) => boolean) | undefined' is not assignable to type 'boolean | ((instance: TableInstance<UseSortByColumnProps<{}>>) => boolean) | undefined'.
                  Type '(instance: TableInstance<{}>) => boolean' is not assignable to type 'boolean | ((instance: TableInstance<UseSortByColumnProps<{}>>) => boolean) | undefined'.
                    Type '(instance: TableInstance<{}>) => boolean' is not assignable to type '(instance: TableInstance<UseSortByColumnProps<{}>>) => boolean'.
                      Types of parameters 'instance' and 'instance' are incompatible.
                        Type 'TableInstance<UseSortByColumnProps<{}>>' is not assignable to type 'TableInstance<{}>'.
                          Types of property 'defaultColumn' are incompatible.
                            Type 'Partial<Column<UseSortByColumnProps<{}>>> | undefined' is not assignable to type 'Partial<Column<{}>> | undefined'.
                              Type 'Partial<Column<UseSortByColumnProps<{}>>>' is not assignable to type 'Partial<Column<{}>>'.
                                Types of property 'accessor' are incompatible.
                                  Type 'string | ((originalRow: UseSortByColumnProps<{}>, index: number, sub: { subRows: UseSortByColumnProps<{}>[]; depth: number; data: UseSortByColumnProps<{}>[]; }) => any) | undefined' is not assignable to type 'string | ((originalRow: {}, index: number, sub: { subRows: {}[]; depth: number; data: {}[]; }) => any) | undefined'.
                                    Type '(originalRow: UseSortByColumnProps<{}>, index: number, sub: { subRows: UseSortByColumnProps<{}>[]; depth: number; data: UseSortByColumnProps<{}>[]; }) => any' is not assignable to type 'string | ((originalRow: {}, index: number, sub: { subRows: {}[]; depth: number; data: {}[]; }) => any) | undefined'.
                                      Type '(originalRow: UseSortByColumnProps<{}>, index: number, sub: { subRows: UseSortByColumnProps<{}>[]; depth: number; data: UseSortByColumnProps<{}>[]; }) => any' is not assignable to type '(originalRow: {}, index: number, sub: { subRows: {}[]; depth: number; data: {}[]; }) => any'.ts(2345)
```

moshthepitt:
@p-netm additionally, I feel strange about calling this `ReactTable` for two reasons:

1. This seems like an ANC, or at the very least an OpenSRP flavored table.  Should we name it something more contextual?
2. Just that odd niggling feeling that comes from the fact that there already exists a package named `ReactTable` that we have already used heavily in other projects.

p-netm:
thanks @moshthepitt @cKellyDesign Will look into this.

clients/core/src/components/ReactTable/index.tsx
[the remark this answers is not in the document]
cKellyDesign: > e.g. you can have a a multi dimensional list

@p-netm ah ok cool good thinking, are there cases where `data` could be `FlexObject[]`?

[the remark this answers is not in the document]
p-netm: >  `extend object` to `T`? Versus something like `ReactTableProps<T = object> `?
@cKellyDesign 
I think the differences between the 2 is this:

``` typescript
interface Data<T extends object> {
    data: T[];
}

interface OData<T = object> {
    data: T[];
}

// This fails as expected, since number is not a non-primitive type like {}, []
// error: Type 'number' does not satisfy the constraint 'object'.ts(2344)
const someData: Data<number> = {};

// This passes, we don't want it to pass, since we expect a non-primitive data type here
const otherData: OData<number> = {};
```

clients/core/src/components/ReactTable/index.tsx
p-netm: yeah, Apparently i will have to case the type to unknown here first

clients/core/src/components/ReactTable/index.tsx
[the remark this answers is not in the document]
cKellyDesign: Cool thanks for the breakdown @p-netm!

packages/OpenSRPTable/src/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                                                    ) : (
+                                                        <>&nbsp;&nbsp;&nbsp;↑</>
+                                                    )
+                                                ) : (
+                                                    <>&nbsp;&nbsp;&nbsp;↕</>
+                                                )
+                                            ) : (
+                                                ''
+                                            )}
+                                        </span>
```
cKellyDesign: Nice, I like this :)

packages/OpenSRPTable/src/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                                    </th>
+                                );
+                            })}
+                        </tr>
+                    ))}
+                </thead>
+                <tbody {...getTableBodyProps()}>
```
cKellyDesign: @p-netm is there ever a case where these `get` methods will return something that's not a spreadable object?

p-netm: I am not sure, These are methods defined by react-table itself. AFAIK no, they should not return anything else than that

cKellyDesign: Yeah I noticed their documentation doesn't specify, but I haven't tried looking at the source code or any type declarations. Would it be possible to `get` these before returning in `render`?

p-netm: @cKellyDesign Guaranteed to return a non-empty object as props, see [here](https://github.com/tannerlinsley/react-table/blob/7dc9b8516543e72410e5dc5ed6ebc10c198df45b/src/makeDefaultPluginHooks.js#L1)

p-netm: Nope that's not sufficient enough i think,... ill keep looking.

p-netm:
A currently blocked on some features that are yet to be added to this component.

**Problem Statement:**
The table only supports ordering the data that is already within its own state, however It also needs to be able to do so, for async-ly fetched data. So how we usually create container components is that we fetch some data, that is saved in store and that leads to a rerender of the container. The problem is that since the `OpenSRPTable` is being rendered as a child of the container, when the container component rerenders, the `OpenSRPTable` is most likely going to reinitialize its state and in the process lose information such as what column was ordered. 

I am still investigating and will add this feature; i also wish to unblock other work that is dependent on this, and thus i will add the above feature in a different issue-PR #219

clients/core/src/components/ReactTable/index.tsx
p-netm: done.

packages/OpenSRPTable/src/index.tsx
p-netm: @cKellyDesign From the documentation:
```
getTableBodyProps: Function(?props)

    Required
    This function is used to resolve any props needed for your table body wrapper.
    Custom props may be passed. NOTE: Custom props will override built-in table body props, so be careful!

```
The above 2 pieces of information are the only ones i could quickly find, they both suggest that the `get...Props` resolve to an object

cKellyDesign: Any way to test/confirm locally? If not then let's resolve this convo :)

yarn.lock
```diff
@@ -2192,11 +2192,14 @@
     "@types/papaparse" "^4.5.11"
     papaparse "^4.6.3"
 
+<<<<<<< HEAD
```
moshthepitt: We probably need to remove these

p-netm: yeah will remove this. Thanks

packages/OpenSRPTable/src/index.tsx
p-netm: @cKellyDesign Nothing comes to mind, but i am pretty confindent at the very least this will resolve to an object. since its default implementation will atleast return a prop object with a prop key of role, in this case the value would be `table`

package.json
```diff
@@ -82,14 +85,16 @@
...
     "workspaces": [
         "clients/*",
         "packages/*"
     ],
     "dependencies": {
-        "@opensrp/server-service": "file:packages/opensrp-server-service/",
-        "istanbul-api": "^2.1.6"
+        "opensrp-server-service": "file:packages/opensrp-server-service",
+        "opensrp-table": "file:packages/OpenSRPTable",
+        "opensrp-web": "file:clients/core"
```
moshthepitt: This seems strange to me:

1. Why are we having `file:` dependencies?
2. Why are we not using the server service package that was published to npm?
3. What needs to depend on `clients/core`, and why?
4. `file:packages/OpenSRPTable` should be changed to the published package, no?

packages/OpenSRPTable/README.md
```diff
@@ -0,0 +1,48 @@
+# Opensrp Table
+
+This provides a grid component that can be used to display data on a page.
```
moshthepitt: Should we make it clear why we are calling it a grid component yet its is called OpenSRP Table?

packages/OpenSRPTable/README.md
```diff
@@ -0,0 +1,48 @@
+# Opensrp Table
```
moshthepitt: Please capitalize the SRP in OpenSRP

packages/OpenSRPTable/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@opensrp/opensrp-table",
+    "version": "0.0.1",
+    "types": "dist/types",
+    "description": "A grid component for displaying data",
```
moshthepitt: Lets also look at the usage of grid here

packages/OpenSRPTable/README.md
p-netm: changed this.

Issue #199, opened 2020-02-19
Refactor READMES'


p-netm:
- Refactor opensrp-web readme.
- Add a proper readme for the repository.

PR #215, opened 2020-02-27
Remove redundant readme header

cKellyDesign:
This doesn't have a ticket, I just noticed when looking at the repo readme.


p-netm:
Yeah there is actually an issue to refactor the readme's will just have this as part of the work leading to the resolution of said issue. #199

cKellyDesign:
Thanks @p-netm

Issue #193, opened 2020-02-12
Rewrite location dux module


p-netm:
Since there is a new proposed strategy for handling locations(which i will be updating as soon as its clear), the current location dux module has been rendered obsolete. And its replacement will be in the following steps:
- [ ] initially Revert Adding the location dux. 
- [ ] Rewriting the location dux module, when it has been sufficiently scoped.

PR #223, opened 2020-03-15
Fix type problem

p-netm:
Fix type problem in master. 

see: 2fa238b 

Also removes location Dux ; part of #193


clients/core/src/store/ducks/baseDux/index.ts
```diff
@@ -103,7 +103,8 @@ export type ImmutableObjectState<ObjectType> = ObjectState<ObjectType> &
     SeamlessImmutable.ImmutableObject<ObjectState<ObjectType>>;
 
 /** initial state */
-const initialState = SeamlessImmutable({
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const initialState: ImmutableObjectState<any> = SeamlessImmutable({
```
redoy11: I am skeptical about this any. Can't we use a flex object here?

p-netm:
> @p-netm why there is another issue about this dux file. is it really necessary to remove it or we can modify the file to resolve that issue later.

@prodhan; it also has the type problem that you raised in baseDux

clients/core/src/store/ducks/baseDux/index.ts
p-netm: @redoy11 I thought that would still not clear the error, just wanted to confirm before I posted this.So from this link 2fa238b. A practical case of what i am trying to talk about is, a developer using baseDux could pass create the reducer like 

```typescript
const reducer = reducerFactory<Array<number>>(reducerName);
```
This would be valid since currently `ObjectType` is not constrained by any types, i.e there is not type constraint in place that says sth like, `ObjectType should be a subtype FlexObject`.

Now considering the reducer above; it would mean that the initial state whose structure is an Empty interface `{}` would not respect the interface `Array<number>`. this is what typescript is trying to tell us.

redoy11: Okay.

p-netm:
An issue to investigate and resolve this in a more standard way: #225

Issue #226, opened 2020-03-16
Fix css and readme in opensrp-web package


p-netm:
Tasks: 
- cleanup: Readme has a few errors that need to be fixed:
  - the uesMemo should have a second argument for dependencies that should be watched to control cache
  - Import statement needs to be fixed.
  - Should change this line : ` return <ReactTable {...{ data, tableColumns: useColumns() }} />;`
- copy the the index.css file into the dist folder next to `index.js`.
  - make a note in the Readme.md for the user to add `import '@opensrp/opensrp-table/dist/index.css'`

prodhan29:
@p-netm , what if i directly use the css file like this `import '@opensrp/opensrp-table/src/index.css'`

p-netm:
> @p-netm , what if i directly use the css file like this `import '@opensrp/opensrp-table/src/index.css'`

@prodhan29 We need to include it in the dist folder. Otherwise it will not get downloaded when one does `yarn add`

PR #227, opened 2020-03-16
Update OpenSRPTable  package

prodhan29:
Fix issue #226


packages/OpenSRPTable/README.md
```diff
@@ -23,7 +23,7 @@ yarn add @opensrp/opensrp-table
 ```typescript
 import React from 'react';
 import { OpenSRPTable } from '@opensrp/opensrp-table';
-import '@opensrp/opensrp-table/dist/index.css'
+import '@opensrp/opensrp-table/dist/index.css';
```
moshthepitt: @p-netm curious whether the `/dist/` is absolutely required (I havent ever tested but I think there is a chance you can just do `import '@opensrp/opensrp-table/index.css';` but we need to test of course

p-netm: I can test this.

p-netm: @moshthepitt For some reason using yarn link did not work, so I had to edit the node_modules file. Anyway my findings:
- as long the `dist/index.js` file has a `require('./index.css')` after transpiling `import './index.css'` and there is a `dist/index.css`, you don't need to do anything extra, the component will load the css file by default.
- following the above: If you add your  own `import @opesrp/opensrp-table/dist/index.css`, nothing new happens.
- following one: if you instead use `import @opensrp/opensrp-table/index.css`, this will result in an error.

moshthepitt: @p-netm thanks.

> as long the dist/index.js file has a require('./index.css') after transpiling import './index.css' and there is a dist/index.css, you don't need to do anything extra, the component will load the css file by default.

Looks like what we want, no?

moshthepitt: > For some reason using yarn link did not work

😲

PR #232, opened 2020-03-27
Pagination added to child register page

prodhan29:
- Add pagination component to child-register page

### DIFF
--- a/clients/core/src/containers/pages/Child/List/index.tsx
+++ b/clients/core/src/containers/pages/Child/List/index.tsx
@@ -5,7 +5,7 @@ import { Col, Row } from 'reactstrap';
 import { Store } from 'redux';
 import Loading from '../../../../components/page/Loading';
 import SearchBox from '../../../../components/page/SearchBox';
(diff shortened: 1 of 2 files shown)


clients/core/src/containers/pages/Child/List/index.tsx
```diff
@@ -133,7 +141,30 @@ class ChildList extends React.Component<ChildListProps, ChildListState> {
...
+    getPaginationOptions = (): PaginationProps => {
+        return {
+            onPageChangeHandler: this.onPageChange,
+            pageNeighbors: PAGINATION_NEIGBOURS,
+            pageSize: PAGINATION_SIZE,
+            totalRecords: this.props.totalRecords,
+        };
+    };
+
+    public render(): ReturnType<React.FC> {
```
p-netm, line 167: @prodhan29 Please also refactor this return type.

prodhan29: it was updated.

PR #233, opened 2020-03-27
ANC Register page

prodhan29:
it contains the below features
- anc data with opensrp table
-  filter option only with searchText
- pagination
- test cases

### DIFF
--- a/clients/core/src/constants.tsx
+++ b/clients/core/src/constants.tsx
@@ -1,5 +1,5 @@
 // endpoints
-export const OPENSRP_CLIENT_ENDPOINT = 'client/search';
+export const OPENSRP_CLIENT_ENDPOINT = 'client/searchByCriteria';
 export type OPENSRP_CLIENT_ENDPOINT = typeof OPENSRP_CLIENT_ENDPOINT;
 
 // internal urls
@@ -13,8 +13,10 @@ export const CLIENT_URL = '/clients';
 export type CLIENT_URL = typeof CLIENT_URL;
 export const HOUSEHOLD_URL = '/404';
 export type HOUSEHOLD_URL = typeof HOUSEHOLD_URL;
-export const ANC_URL = '/404';
+export const ANC_URL = '/anc';
 export type ANC_URL = typeof ANC_URL;
+export const ANC_PROFILE_URL = '/anc-profile';
+export type ANC_PROFILE_URL = typeof ANC_PROFILE_URL;
 export const CHILD_URL = '/child';
 export type CHILD_URL = typeof CHILD_URL;
 export const CHILD_PROFILE_URL = '/child-profile';
@@ -64,6 +66,8 @@ export const SIGN_OUT = 'Sign Out';
 export type SIGN_OUT = typeof SIGN_OUT;
 export const CHILD_CLIENT_TYPE = 'child';
 export type CHILD_CLIENT_TYPE = typeof CHILD_CLIENT_TYPE;
+export const ANC_CLIENT_TYPE = 'anc';
+export type ANC_CLIENT_TYPE = typeof ANC_CLIENT_TYPE;
 
 export const NEXT = 'Next';
 export type NEXT = typeof NEXT;
--- a/clients/core/src/containers/pages/ANC/List/helpers/tableDefinition.tsx
+++ b/clients/core/src/containers/pages/ANC/List/helpers/tableDefinition.tsx
@@ -1,0 +1,71 @@
+/* eslint-disable react/prop-types */
+import React, { useMemo } from 'react';
+import { Link } from 'react-router-dom';
+import { UseTableCellProps } from 'react-table';
+import { ANCClientType } from '../../../../../store/ducks/anc';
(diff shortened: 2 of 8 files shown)


prodhan29:
snapshot of ANC

Screenshot from 2020-03-29 14-21-18

clients/core/src/constants.tsx
```diff
@@ -76,3 +80,9 @@ export type PREVIOUS = typeof PREVIOUS;
 export const LOADING = 'Loading';
 export const VIEW = 'View';
 export type VIEW = typeof VIEW;
+
+/** pagination constants */
+export const PAGINATION_SIZE = 10;
+export type PAGINATION_SIZE = typeof PAGINATION_SIZE;
+export const PAGINATION_NEIGBOURS = 2;
+export type PAGINATION_NEIGBOURS = typeof PAGINATION_NEIGBOURS;
```
moshthepitt, line 88: These should ideally be environment variables, no?

clients/core/src/containers/pages/ANC/List/helpers/tableDefinition.tsx
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable react/prop-types */
```
moshthepitt, line 1: Let us remove magic strings from this module.

clients/core/src/containers/pages/ANC/List/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,50 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
moshthepitt, line 1: This snapshot does not really test anything about the component.  Just shows that the Loading component is displayed.

clients/core/src/constants.tsx
prodhan29: @moshthepitt i moved the endpoints from enviroment to constants file, so as we are moving things from there, i though should also put these variables here.

clients/core/src/containers/pages/ANC/List/helpers/tableDefinition.tsx
prodhan29: okay

clients/core/src/containers/pages/ANC/List/tests/__snapshots__/index.test.tsx.snap
prodhan29: @moshthepitt , i follow this case, which is in master, [sample](https://github.com/OpenSRP/opensrp-web/blob/8cb49fd64ce491a9b3b075b32fc2252d97034d2b/clients/core/src/containers/Clients/List/tests/index.test.tsx#L47)

moshthepitt: @prodhan29 it seems master might also need to be updated

prodhan29: @moshthepitt i updated the test cases

clients/core/src/constants.tsx
prodhan29: these variables has been transferred to the environment file

clients/core/src/configs/env.ts
```diff
@@ -117,5 +117,8 @@ export type OPENSRP_CLIENT_ENDPOINT = typeof OPENSRP_CLIENT_ENDPOINT;
 export const OPENSRP_CHILD_ENDPOINT = process.env.REACT_APP_OPENSRP_CLIENT_ENDPOINT || 'client/searchByCriteria';
 export type OPENSRP_CHILD_ENDPOINT = typeof OPENSRP_CHILD_ENDPOINT;
 
-/** page size for every list of */
-export const PAGINATION_SIZE = 50;
+/** pagination constants */
+export const PAGINATION_SIZE = 10;
+export type PAGINATION_SIZE = typeof PAGINATION_SIZE;
+export const PAGINATION_NEIGBOURS = 2;
+export type PAGINATION_NEIGBOURS = typeof PAGINATION_NEIGBOURS;
```
moshthepitt, line 124: Let us try and get the values from environment variables, with defaults in case not found.

clients/core/src/containers/pages/ANC/List/tests/__snapshots__/index.test.tsx.snap
moshthepitt: You seemed to have removed snapshot tests.  Please fix instead of removing.

prodhan29: @moshthepitt this snapshot was auto-generated because of this [test case](https://github.com/OpenSRP/opensrp-web/pull/233/files#diff-c1a123358f3f2cbef84a15745939e515R42) . i achieved the same thing by checking the length of the ripple. so i removed the snapshot. please correct me if i am wrong.

clients/core/src/configs/env.ts
prodhan29: Done!

PR #254, opened 2020-06-10
Extract reducer factory to a package

ciremusyoka:
Extract reducer factory from core client directory to a package.
Part of https://github.com/onaio/reveal-frontend/issues/900


p-netm:
@ciremusyoka can we get the codacy review passing.

Issue #236, opened 2020-04-02
Create UI for population characteristics editing


ciremusyoka:
The UI is based on [this mock ups ](https://docs.google.com/document/d/1WJGmhKOJiPnoX9GfvnZu4HEw-oFwYooRzEi5nZ5YvLs/edit#)(pg 7,8,9) and the functionality is also as described.

ciremusyoka:
@moshthepitt @dubdabasoduba @p-netm below are images of what i have
![one](https://user-images.githubusercontent.com/26719807/78770082-d1322080-7996-11ea-8596-acb783f838f8.png)
![two](https://user-images.githubusercontent.com/26719807/78770122-dc854c00-7996-11ea-9cf8-1a36acfbfce6.png)
![three](https://user-images.githubusercontent.com/26719807/78770146-e1e29680-7996-11ea-964c-45d17d826ed8.png)

rowo:
@dubdabasoduba hey just wondering, table stylings I thought were already made for OpenSRP web so why not just use the same table/button CSS?

ciremusyoka:
@rowo made some changes on the table text size and stripes color but will rectify to use the OpenSRP web generic ones.

Naima-Bashir:
Tested the UI for this issue. It looks similar to the mockup design: see images below:
![image](https://user-images.githubusercontent.com/61143692/86573853-a0ee2400-bf7d-11ea-9081-f41aa8b950a1.png)
![image](https://user-images.githubusercontent.com/61143692/86573882-aba8b900-bf7d-11ea-93ed-29b36c4ca3ed.png)
POINT TO NOTE:
**currently there is no location/population characteristics. The data is not populated from the server.**

PR #265, opened 2020-07-02
List and edit population characteristics

ciremusyoka:
UI for traversing through user location tree while listing and editing population characteristics settings.
Part of issue #236


ciremusyoka:
screen shots
Screenshot from 2020-07-02 19-13-30
Screenshot from 2020-07-02 19-13-58

packages/populationCharacteristics/package.json
```diff
@@ -0,0 +1,52 @@
+{
+    "name": "@opensrp/population-characteristics",
+    "version": "0.0.3",
```
moshthepitt, line 3: Why is this version 3?  Should it be 0.0.0?

packages/populationCharacteristics/src/components/EditSettings/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+            .then(res => {
+                const { map } = res.locations.locationsHierarchy;
+                const locId = Object.keys(map)[0];
+                getLocations(locId);
+            })
+            .catch(error => {
+                setLoading(false);
+                customAlert && customAlert(String(error), { type: 'error' });
+            });
+    };
```
moshthepitt, line 162: It seems to me that these should all load before the page works.  Instead of what you have here, I would use Promise.all i.e. something like:

```
const promise1 = clientService.list() /// you can do a .then and a .catch if you want
const promise2 = locationService.list() /// you can do a .then and a .catch if you want
Promise.all([promise1, promise2]).then().catch().finally(setLoading(false))
```

You can explore this idea here: https://github.com/onaio/reveal-frontend/blob/f2d44380c92652aa996d3d08b0d165bfb9555e4f/src/containers/pages/PlanAssignment/index.tsx#L184

packages/populationCharacteristics/src/components/EditSettings/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+    // construct table data and headers
+    const listViewProps = {
+        data: locSettings.map(row => {
+            const value = typeof row.value === 'string' ? row.value === 'true' : row.value;
+            return [
+                row.label,
+                row.description,
+                <p key={row.key}>{value ? 'Yes' : 'No'}</p>,
+                row.inheritedFrom?.trim() || '_',
+                <div className="popup" key={row.key}>
```
moshthepitt, line 300: Can we build this component in a helper/util and then use it here by giving it props?

packages/populationCharacteristics/src/components/SearchForm/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+/** function type for custom onChangeHandler functions */
+export type OnChangeType = (event: ChangeEvent<HTMLInputElement>) => void;
+
+/** search bar props interface */
+export interface SearchBarProps extends SearchBarDefaultProps {
+    onChangeHandler: OnChangeType;
+}
+
+/** simple component for handling search */
+const SearchForm = (props: SearchBarProps) => {
```
moshthepitt, line 22: I am concerned that we will end up with too many "SearchForm" components. i.e. this one, the one in the configs package etc.  Can we not have a reusable SearchForm?

packages/populationCharacteristics/src/ducks/locations/index.ts
```diff
@@ -0,0 +1,198 @@
+import { AnyAction, Store } from 'redux';
```
moshthepitt, line 1: Is there a reason why we are not using reducer factory here?

packages/populationCharacteristics/src/ducks/settings/index.ts
```diff
@@ -0,0 +1,121 @@
+import { AnyAction, Store } from 'redux';
```
moshthepitt, line 1: Why are we not using reducerFactory?

packages/populationCharacteristics/src/components/SearchForm/index.tsx
ciremusyoka: have created an issue #266  to extract as a package and assigned myself

ciremusyoka:
@moshthepitt i didn't use reducer factory because the structure i wanted to store the data with is not similar and formatting the data to fit reducer factory format would have taken same effort and amount of code.

packages/populationCharacteristics/src/components/EditSettings/index.tsx
```diff
@@ -0,0 +1,371 @@
+import ListView from '@onaio/list-view';
+import { OpenSRPService } from '@opensrp/server-service';
+import React, { useEffect, MouseEvent, useState, ChangeEvent } from 'react';
+import '../../styles/index.css';
```
p-netm: Maybe best not to bundle the styles this way, IMO it would be better to add them to the dist and have the user import them from there of supply their own.

packages/populationCharacteristics/src/components/EditSettings/index.tsx
```diff
@@ -0,0 +1,371 @@
...
+    getLocDetails,
+    LocChildren,
+    LocPayload,
+    getDefaultLocId,
+} from '../../ducks/locations';
+import { LocationMenu } from '../LocationsMenu';
+import { FormConfigProps, EditSettingLabels } from '../../helpers/types';
+import { SearchForm } from '../SearchForm';
+import { preparePutData, labels } from '../../helpers/utils';
+import '../../../../OpenSRPTable/src/index.css';
```
p-netm: Maybe best not to bundle the styles this way, IMO it would be better to add them to the dist and have the user import them from there of supply their own.

packages/populationCharacteristics/src/components/EditSettings/index.tsx
ciremusyoka: @moshthepitt am not sure if this will be possible because the second promise depends on a response of first.

packages/populationCharacteristics/src/components/LocationsMenu/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+interface DefaultProps {
+    locationDetails: LocChildren;
+    activeLocationId: string;
+    showLocPopUp: string;
+    popLocPopup: (e: MouseEvent, id: string) => any;
+    loadLocsettings: (e: MouseEvent, activeLocId: string) => any;
+    isLast: boolean;
+}
+
```
p-netm, line 12: Lets add docstrings where missing

packages/populationCharacteristics/src/ducks/locations/index.ts
```diff
@@ -0,0 +1,198 @@
...
+
+/** locationsHierarchy parentChildren  interface */
+export interface ParentChildren {
+    [key: string]: Array<string>;
+}
+
+/** locationsHierarchy map interface */
+export interface LocMap {
+    [key: string]: LocChildren;
+}
```
p-netm, line 46: NITPICK:  with onaio/utils; you can know say:

`type ParentChildren = Dictionary<string[]>`
`type locMap = Dictionary<LocChildren>`

as an alternative

packages/populationCharacteristics/src/components/LocationsMenu/index.tsx
ciremusyoka: Docs added

packages/populationCharacteristics/README.md
```diff
@@ -0,0 +1,46 @@
...
+&nbsp;&nbsp; _inheritedLable_
+&nbsp;&nbsp; _inheritSettingsLabel_
+&nbsp;&nbsp; _nameLabel_
+&nbsp;&nbsp; _noDataFound_
+&nbsp;&nbsp; _pageTitle_
+&nbsp;&nbsp; _placeholder_
+&nbsp;&nbsp; _settingLabel_
+&nbsp;&nbsp; _setToNoLabel_
+&nbsp;&nbsp; _setToYesLabel_
+All of the `label` keys expect values of type strings .
```
p-netm, line 46: @ciremusyoka it would be nice to have the code example you had earlier in the readme.

packages/populationCharacteristics/src/components/EditSettings/index.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+    // toggle settings update modal
+    const openEditModal = (e: MouseEvent, row: Setting) => {
+        e.preventDefault();
+        const activeLoc = activeLocationId;
+        fetchSettings([{ ...row, editing: !row.editing }], activeLoc);
+    };
+
+    // update setting
+    const changeSetting = async (e: MouseEvent, row: any, value: string) => {
```
p-netm, line 159: there is an any here that looks like it should be `Setting`, not sure though

packages/populationCharacteristics/src/components/EditSettings/tests/index.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+            wrapper
+                .find('.show div')
+                .at(1)
+                .find('.check').length,
+        ).toEqual(1);
+        // click set to yes
+        wrapper
+            .find('.show div')
+            .at(0)
+            .simulate('click');
```
moshthepitt, line 119: Can we wrap all things that change component state in act? e.g. 

```ts
    act(async () => {
      wrapper
            .find('.show div')
            .at(0)
            .simulate('click');
    });
```

packages/populationCharacteristics/src/components/EditSettings/tests/index.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+        await flushPromises();
+        wrapper.update();
```
moshthepitt, line 121: Can we wrap this in act?  e.g.

```ts
    await act(async () => {
      await flushPromises();
      wrapper.update();
    });
```

PR #288, opened 2020-08-21
Show Custom error when uploading manifest files

### LINKED ISSUE #1119: Manifest upload functionality does not work
https://web.reveal-stage.smartregister.org/manifest/file-upload

image

ciremusyoka:
Fix: https://github.com/onaio/reveal-frontend/issues/1119
Show exact errors if available when user is uploading files instead of showing a general error message.


kelvin-muchiri (dismissed):

packages/opensrp-form-config/src/helpers/services.ts
```diff
@@ -31,8 +31,18 @@ export class OpenSRPServiceExtend extends OpenSRPService {
             if (response.ok || response.status === 201) {
                 return {};
             }
-            const defaultMessage = `OpenSRPService create on ${this.endpoint} failed, HTTP status ${response?.status}`;
-            await throwHTTPError(response, defaultMessage);
+            const responseClone = response.clone();
+            const status = `HTTP status ${response?.status}`;
+            let defaultMessage = `OpenSRPService create on ${this.endpoint} failed, ${status}`;
+            if (!response.ok || response.text) {
```
peterMuriuki: Question :might be wrong, but does response have a text property, I know it has a `text()`

ciremusyoka: Have remove the text check. Have noticed it will always resolve to true since it is always present on a response body.

PR #289, opened 2020-08-25
280 design qa

### LINKED ISSUE #280: Design qa
- [ ] Breadcrumb nav - is this consistent with how this works/looks elsewhere? 
  Screen Shot 2020-08-17 at 10 42 06 PM
- [x] Last item in the breadcrumb should be blue since it's a link
  Screen Shot 2020-08-17 at 10 46 22 PM
- [x] These things should close when something is selected or if you click off. 
  Screen Shot 2020-08-17 at 10 41 57 PM
- [x] This question icon should show a tooltip to explain how inheriting works. Someone can write that up, but it's generally that the higher level defines areas below if all areas below are set to "inherit setting", but if the child is changed, it takes precedence.
  Screen Shot 2020-08-17 at 10 43 25 PM
- [ ] Any reason why we can't do sortable columns on the table like seen on the other admin tables?
  Screen Shot 2020-08-17 at 10 45 07 PM
- [x] The inherit setting button isn't working. It should change the row back to inheriting the setting (even if its the same as what the row is on currently, since it's a different setting).
  Screen Shot 2020-08-17 at 10 45 39 PM
- [x] The inherit setting button is what the item should be default selected. But right now it doesn't even show up unless you change the setting. A "yes" inherited setting is not the same as a "yes" setting since if the parent changes, the inherited one will change.
  Screen Shot 2020-08-17 at 10 55 18 PM

kelvin-muchiri:
Closes #280 

Addresses design issues as specified in the issue

### DIFF
--- a/packages/populationCharacteristics/src/components/EditSettings/index.tsx
+++ b/packages/populationCharacteristics/src/components/EditSettings/index.tsx
@@ -21,7 +21,7 @@ import { FormConfigProps, EditSettingLabels, SettingValue } from '../../helpers/
 import { SearchForm } from '../SearchForm';
 import { labels, EditSettingsButton } from '../../helpers/utils';
 import { POP_CHARACTERISTICS_PARAM, SETTINGS_TRUE } from '../../constants';
-import { editSetting, getInheritedFromLabel, isInheritedFromValid } from './utils';
+import { editSetting, getInheritedFromLabel } from './utils';
 
 /** reqister search and question mark icons */
 library.add(faSearch, faQuestionCircle);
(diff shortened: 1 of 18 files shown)


moshthepitt:
@kelvin-muchiri could we get some screenshots please?

packages/populationCharacteristics/src/components/EditSettings/index.tsx
```diff
@@ -266,7 +270,11 @@ const EditSetings = (props: FormConfigProps & EditSettingsDefaultProps) => {
     // adds question mark to inherited from header
     const iheritedFrom = (
         <span>
-            {inheritedLable} <FontAwesomeIcon icon="question-circle" />
+            {inheritedLable}{' '}
+            <span className="opensrp-tooltip">
+                <FontAwesomeIcon icon="question-circle" />
+                <span className="tooltiptext">{TOOLTIP_INHERITED_FROM}</span>
```
ciremusyoka, line 276: let's have `TOOLTIP_INHERITED_FROM` assigned and passed as a  prop

